PR #80, opened 2024-05-20
feat: use abigen code for contracts and add debug service

aloknerurkar:
- We added a new library for the oracle to cleanly use the abigen code for making contract calls/transactions. This change will make use of the same library in p2p to use the abigen code. This will reduce a lot of unnecessary code and be less error prone. Also this provides safer sequential transaction processing from the nodes.
- Added a new gRPC service called debug to house common debugging APIs. Previously the transaction related helper APIs were only available on provider node. Now the bidder node also could make transactions. These debugging APIs will be needed again. So having a common service makes more sense. This will also include the topology API so it can be documented and versioned. This can also be useful to try out experimental APIs.
- Oracle allows limited parallel settlements to prioritize the record winner transaction.


p2p/pkg/preconfirmation/tracker/tracker.go
```diff
@@ -222,14 +230,29 @@ func (t *Tracker) handleNewL1Block(
 		}
 		startTime := time.Now()
 
+		var commitmentIdx [32]byte
+		copy(commitmentIdx[:], commitment.CommitmentIndex[:])
+
+		bidAmt, ok := new(big.Int).SetString(commitment.Bid.BidAmount, 10)
```
Mikelle, line 236: No point to make `bidAmt` as `*big.Int`
```suggestion
		bidAmt, err := strconv.ParseUint(commitment.Bid.BidAmount, 10, 64)
 ```

File: p2p/pkg/preconfirmation/tracker/tracker.go [hunk 2]
```diff
@@ -222,14 +230,29 @@ func (t *Tracker) handleNewL1Block(
...
-			commitment.PreConfirmation.Bid.BidAmount,
-			commitment.PreConfirmation.Bid.BlockNumber,
+			opts,
+			commitmentIdx,
+			bidAmt.Uint64(),
+			uint64(commitment.PreConfirmation.Bid.BlockNumber),
 			commitment.PreConfirmation.Bid.TxHash,
-			commitment.PreConfirmation.Bid.DecayStartTimestamp,
-			commitment.PreConfirmation.Bid.DecayEndTimestamp,
+			uint64(commitment.PreConfirmation.Bid.DecayStartTimestamp),
```
Mikelle, line 254: We could change from `int64` to `uint64` in .proto file and avoid redundant conversion

p2p/pkg/store/store.go
```diff
@@ -308,3 +318,41 @@ func (s *Store) Len() int {
...
 	return s.Tree.Len()
 }
+
+type TxnDetails struct {
+	Hash    common.Hash
+	Nonce   uint64
+	Created int64
+}
+
+func (s *Store) Save(ctx context.Context, txHash common.Hash, nonce uint64) error {
```
Mikelle, line 328:
```suggestion
func (s *Store) SaveTxnDetails(ctx context.Context, txHash common.Hash, nonce uint64) error {
```

p2p/pkg/preconfirmation/tracker/tracker.go
Mikelle: This value will overflow because of the conversion further in the code, so converting to bigInt will not help here
`bidAmt.Uint64()`

p2p/pkg/store/store.go
Mikelle: Thank you for your comments around Save and Update. I understand the reasoning behind the naming related to the txmonitor.Saver interface. However, I still believe that renaming the function to SaveTxnDetails would improve clarity.

While the current name might seem logical within the txmonitor package, it causes some confusion within store.go. The lack of logical separation between methods in store.go makes it hard to understand the purpose of the Save method without referring back to the txmonitor package. Renaming it to SaveTxnDetails would make the function's intent clearer within the context of store.go.

Mikelle: Thank you for the detailed comment! It makes sense to me now that the store wasn't supposed to implement this interface.

Mikelle (approved):

PR #126, opened 2024-06-03
refactor: deploy contracts using send raw transaction

kant777:
There seems to be a lot of issues with deploy_contract.job especially related to foundry version etc. While foundry is a great tool I thought deploying contracts using `eth_sendRawTransaction` should be the easiest and cleanest way to do it. 

The python script simply grabs the required contracts and deploys them to ethereum using `send_raw_transaction`

How to run:

1. `pip install web3`
2. make sure you are under `mev-commit/contracts`
3. `forge build` (sometimes `forge build` complains. If so, run it with `forge build --via-ir` )
4. `python3 deploy.py`

### DIFF
--- a/.github/workflows/infrstructure.yml
+++ b/.github/workflows/infrstructure.yml
@@ -117,6 +117,9 @@ jobs:
           check-latest: true
           cache-dependency-path: go.work.sum
 
+      - name: Setup Foundry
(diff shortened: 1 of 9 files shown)


kant777:
I will add deploy_create2.sh in a similar fashion

PR #135, opened 2024-06-04
feat: improve build and security

mrekucci:
- keystores, passwords, and `genesis.json` file are generated at build time and stored in a secure vault
- addresses of deployed contracts are exposed via http
- introduces environments on top of existing profiles
- various enhancements and fixes


external/geth
shaspitz, line 1: This shouldn't change the `period` and `gasLimit` of the chain. Also shouldn't the genesis `alloc` field be a lot smaller now too?

PR #147, opened 2024-06-14
feat: improve validator registry for migration to holesky

shaspitz:


contracts/contracts/ValidatorRegistryV1.sol
```diff
@@ -3,13 +3,15 @@ pragma solidity ^0.8.20;
...
 
-/// @title Validator Registry
-/// @notice Logic contract enabling L1 validators to opt-in to mev-commit via staking. 
+/// @title Validator Registry v1
+/// @notice Logic contract enabling L1 validators to opt-in to mev-commit 
+/// via simply staking ETH outside what's staked with the beacon chain.
 /// @dev Slashing is not yet implemented for this contract, hence it is upgradable to incorporate slashing in the future.
-/// @dev This contract is meant to be deployed via a proxy contract.
-contract ValidatorRegistry is OwnableUpgradeable, ReentrancyGuardUpgradeable {
+/// @dev This contract is meant to be deployed via UUPS proxy contract on mainnet.
```
ckartik, line 13: Are we deploying to holesky or mainnet?

contracts/entrypoint.sh
```diff
@@ -140,5 +137,9 @@ elif [ "${DEPLOY_TYPE}" = "validator-registry" ]; then
         -vvvv \
         --use 0.8.23 \
         --root "${CONTRACT_REPO_ROOT_PATH}" \
-        --via-ir
+        --via-ir \
+        --skip-simulation \
+        --force \
```
ckartik, line 142: Weird that they use force to describe clearing a cache, but good idea to keep it here!

Issue #153, opened 2024-06-17
Change from uint64 to uint256 for the bid amount


Mikelle:
Uint64 is too small for the bid amount; it will allow bids only up to approximately 16 ETH.

PR #157, opened 2024-06-17
chore: moved from uint64 to uint256

Mikelle:
https://github.com/primev/mev-commit/issues/153


oracle/pkg/updater/updater_test.go
```diff
@@ -285,7 +287,7 @@ func TestUpdater(t *testing.T) {
 			if !bytes.Equal(settlement.builder, c.Commiter.Bytes()) {
 				t.Fatal("wrong builder")
 			}
-			if settlement.amount != 0 {
+			if settlement.amount.Uint64() != 10 {
```
ckartik, line 290: why are you getting the Uint64 value here?

ckartik (approved):

oracle/pkg/updater/updater_test.go
Mikelle: Thank you for the review! This is for the test. The initial value is 10, so it can be converted to uint64 here.

PR #169, opened 2024-06-24
feat: productionize validator registry v1

shaspitz:
- Implements slashing
- Adds setters for changing onchain params, only callable by owner
- Contract now uses modifiers
- Makes contract pauseable
- Consolidates multiple mappings into a single mapping w/ struct values, simplifying logic
- Adds registry interface
- Splits contract into state and state transition functionality, as is best practice for upgradable contracts
- Adds comments + readme


shaspitz:
Thanks @Mikelle! Your comments should have been addressed in https://github.com/primev/mev-commit/commit/44c2bd86ba7394a177569e37d2c1dba5a4a23273

Mikelle (approved):

PR #171, opened 2024-06-25
Adds reverting Txn list

ckartik:
This PR updates the PreConfCommitmentStore smart contract by adding support for handling transactions that are can  revert. It introduces a new field, revertingTxHashes, to track these transactions, allowing the system to process them without affecting the integrity of the entire transaction bundle.


p2p/pkg/rpc/bidder/service.go
```diff
@@ -99,6 +99,7 @@ func (s *Service) SendBid(
 	}
 
 	txnsStr := strings.Join(bid.TxHashes, ",")
+	revertingTxHashesStr := strings.Join(bid.RevertingTxHashes, ",")
```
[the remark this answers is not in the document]
ckartik: Can we keep that as a seperate PR, we do that for the txnsstr as well here, would require more refactor.

ckartik:
Stress testing running well [here](https://app.datadoghq.com/dashboard/p65-un4-h4x/mev-commit-oracle?fromUser=true&refresh_mode=sliding&tpl_var_version%5B0%5D=d4193dc-1720549161&view=spans&from_ts=1720551086754&to_ts=1720551386754&live=true)

p2p/pkg/signer/preconfencryptor/encryptor.go
```diff
@@ -348,19 +353,22 @@ func GetPreConfirmationHash(c *preconfpb.PreConfirmation) ([]byte, error) {
...
-		[]byte("PreConfCommitment(string txnHash,uint256 bid,uint64 blockNumber,uint64 decayStartTimeStamp,uint64 decayEndTimeStamp,bytes32 bidHash,string signature,string sharedSecretKey)"),
+		[]byte("PreConfCommitment(string txnHash,string revertingTxHashes,uint256 bid,uint64 blockNumber,uint64 decayStartTimeStamp,uint64 decayEndTimeStamp,bytes32 bidHash,string signature,string sharedSecretKey)"),
 	)
 
 	// Convert the txnHash to a byte array and hash it
 	txnHashHash := crypto.Keccak256Hash([]byte(c.Bid.TxHash))
+	revertingTxHashesHash := crypto.Keccak256Hash([]byte(c.Bid.RevertingTxHashes))
 	bidDigestHash := crypto.Keccak256Hash([]byte(hex.EncodeToString(c.Bid.Digest)))
 	bidSigHash := crypto.Keccak256Hash([]byte(hex.EncodeToString(c.Bid.Signature)))
+	fmt.Printf("Bid Signature Hash: %x\n", bidSigHash.Bytes())
```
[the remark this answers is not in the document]
ckartik: Sorry, that was a debug log.

contracts/contracts/PreConfCommitmentStore.sol
```diff
@@ -94,13 +95,28 @@ contract PreConfCommitmentStore is OwnableUpgradeable, UUPSUpgradeable {
...
+        string revertingTxHashes,
         bytes32 commitmentHash,
         bytes bidSignature,
         bytes commitmentSignature,
         uint64 dispatchTimestamp,
         bytes sharedSecretKey
     );
 
+    /// @dev Struct for all the commitment params to avoid too deep in the stack error
+    struct CommitmentParams {
```
shaspitz, line 107: The fields in `PreConfCommitment` that are duplicated in `CommitmentParams` should be removed from `PreConfCommitment`, or we can add docs on why the repeated fields need to separately exist in both structs

contracts/contracts/PreConfCommitmentStore.sol
```diff
@@ -308,38 +332,45 @@ contract PreConfCommitmentStore is OwnableUpgradeable, UUPSUpgradeable {
...
+    }
+
+    function _verifyPreConfCommitment(
+        CommitmentParams memory params
+    ) internal view returns (bytes32 preConfHash, address commiterAddress) {
+        preConfHash = _getPreConfHash(params);
+        commiterAddress = _recoverCommitterAddress(preConfHash, params.commitmentSignature);
+    }
+
+    function _getPreConfHash(
```
shaspitz, line 353: The public function `getPreConfHash` should instead call the internal `_getPreConfHash` function, where underlying functionality lives in the internal func

contracts/contracts/PreConfCommitmentStore.sol
```diff
@@ -394,6 +425,7 @@ contract PreConfCommitmentStore is OwnableUpgradeable, UUPSUpgradeable {
         uint256 bid,
         uint64 blockNumber,
         string memory txnHash,
+        string memory revertingTxHashes,
```
shaspitz, line 428: nit add `@param` comment for revertingTxHashes

contracts/contracts/PreConfCommitmentStore.sol
```diff
@@ -294,10 +316,12 @@ contract PreConfCommitmentStore is OwnableUpgradeable, UUPSUpgradeable {
         uint64 decayStartTimeStamp,
         uint64 decayEndTimeStamp,
         string memory txnHash,
+        string memory revertingTxHashes,
```
shaspitz, line 319: Nit add `@param` comment for revertingTxHashes

contracts/contracts/PreConfCommitmentStore.sol
```diff
@@ -250,6 +269,7 @@ contract PreConfCommitmentStore is OwnableUpgradeable, UUPSUpgradeable {
      */
     function getPreConfHash(
         string memory _txnHash,
+        string memory _revertingTxHashes,
```
shaspitz, line 272: Nit add `@param` comment for _revertingTxHashes

contracts/contracts/PreConfCommitmentStore.sol
```diff
@@ -219,25 +236,27 @@ contract PreConfCommitmentStore is OwnableUpgradeable, UUPSUpgradeable {
      */
     function getBidHash(
         string memory _txnHash,
+        string memory _revertingTxHashes,
```
shaspitz, line 239: Nit add `@param` comment for revertingTxHashes

contracts/contracts/PreConfCommitmentStore.sol
```diff
@@ -19,15 +19,15 @@ contract PreConfCommitmentStore is OwnableUpgradeable, UUPSUpgradeable {
 
     /// @dev EIP-712 Type Hash for preconfirmation commitment
     bytes32 public constant EIP712_COMMITMENT_TYPEHASH =
```
shaspitz, line 21: revertingTxHashes should always be a csv string right? If so we should document this in contracts or add format validation  

Also it seems that revertingTxHashes represent multiple allowed reverting tx hashes. Wouldn't it only be relevant if `txnHash` for a preconf reverted? Hence we'd only need to post a boolean

oracle/pkg/updater/updater.go
```diff
@@ -336,19 +336,30 @@ func (u *Updater) handleOpenedCommitment(
...
+	revertableTxnsMap := make(map[string]bool)
+	for _, txn := range revertableTxns {
+		revertableTxnsMap[txn] = true
+	}
+
 	// Ensure Bundle is atomic and present in the block
 	for i := 0; i < len(commitmentTxnHashes); i++ {
 		txnDetails, found := txns[commitmentTxnHashes[i]]
-		if !found || txnDetails.PosInBlock != (txns[commitmentTxnHashes[0]].PosInBlock)+i || !txnDetails.Succeeded {
+		if !found || txnDetails.PosInBlock != (txns[commitmentTxnHashes[0]].PosInBlock)+i || (!txnDetails.Succeeded && !revertableTxnsMap[commitmentTxnHashes[i]]) {
```
shaspitz, line 352: This is a lot of logic to pack into a single line, consider using multiple ifs for readability

contracts/contracts/PreConfCommitmentStore.sol
shaspitz: my second question may have to do with a lack of understanding of the current preconf structure

shaspitz:
Prob too late to change, but `revertingTxHashes` semantically sounds like txes that _are reverting_ rather than intended meaning of _txes that are allowed to revert_. (`revertAllowedTxHashes` possible name)

ckartik:
@shaspitz The term comes from builder RPC endpoints, where they specify it as `revertingTxHashes`

oracle/pkg/updater/updater.go
ckartik: Going to keep this for now, because I had it split into multiple Ifs and other reviewers didn't like it.

contracts/contracts/PreConfCommitmentStore.sol
ckartik: It's the same as the txnhashes array. There can be mutliple txn hashes in a single bid

contracts/contracts/PreConfCommitmentStore.sol
ckartik: I think it's fine to keep both, the reason _getPreConfHash was added iirc, is because the yul compiler couldnt handle us passing in varaibles one by one on the stack. Not sure the rationale here, Kant helped fix this one, and I don't want to mess with it too much in fear of reintroducing the same issue.

contracts/contracts/PreConfCommitmentStore.sol
ckartik: Not sure I gather what you're referring to here.

shaspitz: All of these fields are already a part of `PreConfCommitment` and repeated code is not ideal

contracts/contracts/PreConfCommitmentStore.sol
shaspitz: No logic changes are needed, I'm just saying that common solidity pattern is that public functions call internal functions which hold the underlying logic. See `ValidatorRegistryV1.sol` as example

shaspitz:
> @shaspitz The term comes from builder RPC endpoints, where they specify it as `revertingTxHashes`

Endpoints we designed? Or endpoints external to our protocol

contracts/contracts/PreConfCommitmentStore.sol
shaspitz: This has to do with different gas costs for public/private/internal/external

ckartik:
Endpoints that are part of the PBS pipeline

shaspitz (approved):

contracts/contracts/PreConfCommitmentStore.sol
ckartik: Will make this change later. I haven't stress tested the changed approach.

PR #174, opened 2024-06-26
fix: forge build now specifies correct script directory 

shaspitz:
Previously bugs like https://github.com/primev/mev-commit/pull/173 would not be caught by ci, because forge thinks scripts reside in the `script` directory by default. Our script directory is actually `scripts` as addressed in this PR. Now `forge test`, `forge build` etc. will fail as expected when one of the solidity scripts doesn't build.

### DIFF
--- a/contracts/scripts/DeployScripts.s.sol
+++ b/contracts/scripts/DeployScripts.s.sol
@@ -6,7 +6,6 @@ import "../contracts/ProviderRegistry.sol";
 import "../contracts/PreConfCommitmentStore.sol";
 import "../contracts/Oracle.sol";
 import "../contracts/Whitelist.sol";
-import "../contracts/validator-registry/ValidatorRegistryV1.sol";
 import {Upgrades} from "openzeppelin-foundry-upgrades/Upgrades.sol";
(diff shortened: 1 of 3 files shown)


contracts/scripts/DeployScripts.s.sol
```diff
@@ -6,7 +6,6 @@ import "../contracts/ProviderRegistry.sol";
 import "../contracts/PreConfCommitmentStore.sol";
 import "../contracts/Oracle.sol";
 import "../contracts/Whitelist.sol";
-import "../contracts/validator-registry/ValidatorRegistryV1.sol";
```
shaspitz, line 9: this dep is unneeded for core deployment

contracts/scripts/DeployStandardBridge.s.sol
```diff
@@ -1,35 +1,38 @@
 // SPDX-License-Identifier: BSL 1.1
```
shaspitz, line 1: Changes necessary to make this script build and pass ci

kant777 (approved):

PR #175, opened 2024-06-27
feat: mev-commit AVS for validator opt-in

shaspitz:
- Adds audit-ready AVS contracts to enable validators to restake to opt into our protocol
- Eigenlayer contracts are now a solidity dep
- Adds static png and json for eigenlayer frontend integration
- Adds router contract which allows actor to query if a validator is opted in to either our restaking or simple staking solution
- Updates readmes w.r.t new changes
- Some tests, more to come.


shaspitz:
closing in favor of https://github.com/primev/mev-commit/pull/176 since one of my commits does not pass naming ci

PR #188, opened 2024-07-02
feat: add documentation details around keys

ckartik:


mrekucci:
This is already documented [here](https://github.com/primev/mev-commit/blob/main/infrastructure/nomad/README.md?plain=1#L120).

PR #211, opened 2024-07-08
feat: autodeposit bidder balance

aloknerurkar:
This PR adds a new autodeposit feature to the Bidder API. With this new API, users can configure a minimum balance per window that the node will deposit into the bidder registry. As and when the windows move ahead, the node will withdraw the old deposits and make new ones for the new window.

All refunds will be made to the bidder's wallet. These include return of decayed amount as well as bid amounts in case of slashes.


p2p/pkg/rpc/bidder/service.go
```diff
@@ -316,3 +319,204 @@ func (s *Service) Withdraw(
...
+			for range ticker.C {
+				currentWindow, err := s.blockTrackerContract.GetCurrentWindow()
+				if err != nil {
+					s.logger.Error("getting current window", "error", err)
+					continue
+				}
+				doWithdraw := true
+				for _, w := range windows {
+					if w.Uint64() >= currentWindow.Uint64() {
+						doWithdraw = false
```
Mikelle, line 392: It would probably be good to add a log here indicating that we are waiting for the correct window, so it will be clearer to the user what is happening

Mikelle (approved):

PR #214, opened 2024-07-08
fix: providers specify BLS pubkey during registration

shaspitz:


Mikelle (approved):

p2p/integrationtest/provider/client.go
```diff
@@ -123,6 +124,12 @@ func (b *ProviderClient) CheckAndStake() error {
 		BlsPublicKey: hex.EncodeToString(blsPubkeyBytes),
 	})
 	if err != nil {
+		if strings.Contains(err.Error(), "Invalid BLS public key length") {
+			b.logger.Error("failed to register stake", "err", err)
```
Mikelle, line 128: Could we do the logging in one line?

shaspitz: Yea I was just using this log for temp debugging, so it's removed in https://github.com/primev/mev-commit/commit/ed2f54befc0e33e4e94d813560a033bcdb587d51

PR #229, opened 2024-07-11
feat: add new API for provider registration

aloknerurkar:
Adds password protected API for registering providers with the block tracker contract.


oracle/pkg/apiserver/apiserver.go
```diff
@@ -71,10 +86,81 @@ func New(
...
 }
 
+func (s *Service) registerProvider(password string) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.Method != http.MethodPost {
+			http.Error(w, "Method not allowed", http.StatusMethodNotAllowed)
+			return
+		}
+
+		if r.Header.Get("X-Registration-Password") != password {
```
Mikelle, line 102: How do we plan to use this password? I don’t see that it is saved or used later in the code.

Mikelle (approved):

PR #230, opened 2024-07-11
chore: standardize solidity version to 0.8.20 across repo

shaspitz:
See https://swcregistry.io/docs/SWC-103/. This branch [looks good in devnet setup](https://github.com/primev/mev-commit/actions/runs/9900092362) and should effectively be a refactor. Ie. we've been using 0.8.20 already but this PR makes it explicit everywhere


Mikelle:
Why did we decide to use 0.8.20 and not 0.8.26, for example?

shaspitz:
> Why did we decide to use 0.8.20 and not 0.8.26, for example?

If I remember correctly it first had to do with matching the version of one of our open-zeppelin dependancies. At this point 0.8.20 is just what we've tested the most and designed around. So no need to upgrade until a future solidity version has a necessary feature for our uses

Issue #233, opened 2024-07-12
Go through mev-commit smart contracts and do optimizations


Mikelle:
Example:
* i++ -> ++i
* reduce redundant state reading
* reduce messages in require to 32 bytes
* etc...

PR #234, opened 2024-07-12
fix: refactored contracts

Mikelle:
Closes: https://github.com/primev/mev-commit/issues/233


contracts/contracts/PreConfCommitmentStore.sol
```diff
@@ -721,7 +698,6 @@ contract PreConfCommitmentStore is OwnableUpgradeable, UUPSUpgradeable {
     function _bytes32ToHexString(
         bytes32 _bytes32
     ) internal pure returns (string memory) {
-        bytes memory HEXCHARS = "0123456789abcdef";
         bytes memory _string = new bytes(64);
         for (uint8 i = 0; i < 32; i++) {
```
shaspitz, line 702:
```suggestion
        for (uint8 i = 0; i < 32; ++i) {
```

contracts/contracts/PreConfCommitmentStore.sol
```diff
@@ -738,7 +714,6 @@ contract PreConfCommitmentStore is OwnableUpgradeable, UUPSUpgradeable {
     function _bytesToHexString(
         bytes memory _bytes
     ) public pure returns (string memory) {
-        bytes memory HEXCHARS = "0123456789abcdef";
         bytes memory _string = new bytes(_bytes.length * 2);
         for (uint256 i = 0; i < _bytes.length; i++) {
```
shaspitz, line 718:
```suggestion
        for (uint256 i = 0; i < _bytes.length; ++i) {
```

contracts/contracts/PreConfCommitmentStore.sol
```diff
@@ -558,13 +540,13 @@ contract PreConfCommitmentStore is OwnableUpgradeable, UUPSUpgradeable {
         uint256 minTime = block.timestamp - commitmentDispatchWindow;
 
         // Check if the dispatch timestamp is within the allowed dispatch window
-        require(dispatchTimestamp > minTime, "Invalid dispatch timestamp");
+        require(dispatchTimestamp >= minTime, "Invalid dispatch timestamp");
```
shaspitz, line 543: Is this change of behavior intentional? After clarifying with the auditors their stance is `When inside the require statements, non-strict inequalities (>=, <=) are usually costlier than strict equalities (>, <).`

shaspitz (approved):
Looks solid! Just a few comments

contracts/contracts/PreConfCommitmentStore.sol
Mikelle: I did it, after reading this quote:
```
When inside the “if/require” statements, non-strict inequalities (>=, <=) are
usually cheaper than the strict equalities (>, <).
```
Thank you for clarifying, I will return it back

Issue #109, opened 2024-05-27
Separate accounts for contract deployer and oracle


shaspitz:
Currently the permissioned oracle account must be the same as the account which deployed the core contracts, see [here](https://github.com/primev/mev-commit/blob/80781bf15e95120dfb871cfee0f24fef5c206799/infrastructure/nomad/playbooks/templates/jobs/mev-commit-oracle.nomad.j2#L193). From a security perspective we should introduce two separate keystores for the oracle and the contract deployer. This would involve changing the onlyOwner modifier for multiple contract functions (ex: [here](https://github.com/primev/mev-commit/blob/80781bf15e95120dfb871cfee0f24fef5c206799/contracts/contracts/Oracle.sol#L80)) s.t. it checks for a separate oracle account who sent the tx

PR #241, opened 2024-07-16
fix: remaining audit remediations

shaspitz:
Fixes issues 6-14 from recent validator registry audit. Most are refactors or gas optimizations. Devnet tested in https://github.com/primev/mev-commit/actions/runs/9996080181


contracts/contracts/PreConfCommitmentStore.sol
```diff
@@ -14,7 +14,7 @@ import {WindowFromBlockNumber} from "./utils/WindowFromBlockNumber.sol";
  * @title PreConfCommitmentStore - A contract for managing preconfirmation commitments and bids.
  * @notice This contract allows bidders to make precommitments and bids and provides a mechanism for the oracle to verify and process them.
  */
-contract PreConfCommitmentStore is OwnableUpgradeable, UUPSUpgradeable {
+contract PreConfCommitmentStore is Ownable2StepUpgradeable, UUPSUpgradeable {
```
Mikelle, line 17: Is a change to `Ownable2StepUpgradeable` needed? If we change the mechanism to transfer ownership, do we need to update this script: https://github.com/primev/mev-commit/blob/main/contracts/scripts/DeployScripts.s.sol#L78?

shaspitz: I wasn't aware we're transferring ownership for only two of our contracts. I'll work on https://github.com/primev/mev-commit/issues/109 as that is the proper solution here

shaspitz: As far as ownable2Step here's what auditors said

> The "Ownable2Step" pattern is an improvement over the traditional "Ownable" pattern, designed to enhance the security of ownership transfer functionality in a smart contract. Unlike the original "Ownable" pattern, where ownership can be transferred directly to a specified address, the "Ownable2Step" pattern introduces an additional step in the ownership transfer process. Ownership transfer only completes when the proposed new owner explicitly accepts the ownership, mitigating the risk of accidental or unintended ownership transfers to mistyped addresses. 

> Impacts
Without the "Ownable2Step" pattern, the contract owner might inadvertently transfer ownership to an unintended or mistyped address, potentially leading to a loss of control over the contract. By adopting the "Ownable2Step" pattern, the smart contract becomes more resilient against external attacks aimed at seizing ownership or manipulating the contract's behavior.

shaspitz: Resolving this comment from https://github.com/primev/mev-commit/pull/251 being merged

Mikelle (approved):

PR #256, opened 2024-07-19
perf: streamline transactions from node

aloknerurkar:
- Use constant gas values to prevent gas (estimate, suggesttip)
- Buffer the expensive events to allow other events to go through
- fixed bug to allow duplicate encrypted commitments to go through

18hr run on stressnet stable

Screenshot from 2024-07-20 13-20-08
Screenshot from 2024-07-20 13-19-37

### DIFF
--- a/oracle/cmd/main.go
+++ b/oracle/cmd/main.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io"
 	"log/slog"
+	"math/big"
 	"os"
 	"path/filepath"
 	"slices"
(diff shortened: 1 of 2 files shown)


x/keysigner/keystoresigner.go
```diff
@@ -77,7 +77,15 @@ func (kss *KeystoreSigner) String() string {
 }
 
 func (kss *KeystoreSigner) GetAuth(chainID *big.Int) (*bind.TransactOpts, error) {
-	return bind.NewKeyStoreTransactorWithChainID(kss.keystore, kss.account, chainID)
+	opts, err := bind.NewKeyStoreTransactorWithChainID(kss.keystore, kss.account, chainID)
+	if err != nil {
+		return nil, err
+	}
+
+	opts.GasLimit = 1_000_000
```
Mikelle, line 85: Wouldn’t it be better to declare those variables as constants or read them from the config?

p2p/pkg/preconfirmation/tracker/tracker.go
```diff
@@ -87,7 +87,7 @@ func NewTracker(
 		preconfContract: preconfContract,
 		receiptGetter:   receiptGetter,
 		optsGetter:      optsGetter,
-		newL1Blocks:     make(chan *blocktracker.BlocktrackerNewL1Block),
+		newL1Blocks:     make(chan *blocktracker.BlocktrackerNewL1Block, 5),
```
Mikelle, line 90: Should we declare those parameters as constants and add a comment explaining why 5 and not 10, for example?

oracle/pkg/store/store.go
```diff
@@ -157,6 +172,11 @@ func (s *Store) AddEncryptedCommitment(
 	commitmentSignature []byte,
 	dispatchTimestamp uint64,
 ) error {
+	// check if the commitmentIdx already exists
+	if s.encryptedCommitmentExists(ctx, commitmentIdx) {
```
Mikelle, line 176: “Wouldn’t it be better to add error handling (e.g., skip if duplicated) instead of checking for the existence of the commitment for every request?”

kant777 (approved):

PR #258, opened 2024-07-19
Fully enable solhint merge

Mikelle:


contracts/contracts/BidderRegistry.sol
```diff
@@ -240,12 +250,10 @@ contract BidderRegistry is
             totalAmount += amount;
         }
 
-        if (totalAmount == 0) {
```
shaspitz, line 243: was it intentional to remove this?

contracts/contracts/BidderRegistry.sol
```diff
@@ -494,6 +519,5 @@ contract BidderRegistry is
         return lockedFunds[bidder][window];
     }
 
-    // solhint-disable-next-line no-empty-blocks
-    function _authorizeUpgrade(address) internal override onlyOwner {}
+    function _authorizeUpgrade(address) internal override onlyOwner {} // solhint-disable no-empty-blocks
```
shaspitz, line 522: fyi `// solhint-disable` disables for the rest of the file, so that's why `// solhint-disable-next-line no-empty-blocks` existed here previously

contracts/contracts/Oracle.sol
```diff
@@ -133,6 +133,8 @@ contract Oracle is Ownable2StepUpgradeable, UUPSUpgradeable {
         emit OracleAccountSet(oldOracleAccount, newOracleAccount);
     }
 
+    function _authorizeUpgrade(address) internal override onlyOwner {} // solhint-disable no-empty-blocks
```
shaspitz, line 136: same as above

contracts/contracts/BlockTracker.sol
```diff
@@ -167,8 +168,10 @@ contract BlockTracker is Ownable2StepUpgradeable, UUPSUpgradeable {
     */
     function _recordBlockWinner(uint256 blockNumber, address winner) internal {
         // Check if the block number is valid (not 0)
-        require(blockNumber != 0, "Invalid block number");
+        if (blockNumber == 0) revert InvalidBlockNumber();
 
         blockWinners[blockNumber] = winner;
     }
+
+    function _authorizeUpgrade(address) internal override onlyOwner {} // solhint-disable-line no-empty-blocks
```
shaspitz, line 176: I think this works because it's `solhint-disable-line`

shaspitz (approved):

contracts/contracts/BidderRegistry.sol
Mikelle: Yes, bcs this check was happening on each function call, it's cheaper to remove it.

Issue #276, opened 2024-07-25
Smart contract changes


Mikelle:
Changes to the contracts logic:

- [x]  In case of double spend restriction, instead of rounding the overflow and left it in the contract, send it back to the bidder
- [x]  Change withdrawal logic for the provider to the two steps:
    ◦ Ask for withdrawal
    ◦ After waiting for the 24 hours, provider will be able to withdraw
- [x] In case provider funds got slashed, send slashed funds directly to the bidder address instead of balance map in provider registry
- [x] When provider is storing commitment, check that his balance is bigger than minStake

PR #279, opened 2024-07-29
feat: smart contract updates

Mikelle:
## Describe your changes

Updating smart contracts (details in the issue)

Fixes https://github.com/primev/mev-commit/issues/276


contracts/contracts/ProviderRegistry.sol
```diff
@@ -194,53 +189,37 @@ contract ProviderRegistry is
...
-    function withdrawStakedAmount(
-        address payable provider
-    ) external nonReentrant {
-        require(msg.sender == provider, "Only provider can unstake");
-        uint256 stake = providerStakes[provider];
-        providerStakes[provider] = 0;
+    /// @dev Withdraws the staked amount after 24 hours of withdrawal request.
+    function withdrawStakedAmount() external nonReentrant {
+        require(withdrawalRequests[msg.sender] > 0, "No withdrawal request");
+        require(block.timestamp >= withdrawalRequests[msg.sender] + 24 hours, "24 hours have not passed");
```
shaspitz, line 202: withdraw period should be configurable with an only-owner callable setter

contracts/contracts/ProviderRegistry.sol
```diff
@@ -194,53 +189,37 @@ contract ProviderRegistry is
...
-     */
-    function withdrawBidderAmount(address bidder) external nonReentrant {
-        require(bidderAmount[bidder] > 0, "Bidder Amount is zero");
-
-        bidderAmount[bidder] = 0;
-
-        (bool success, ) = bidder.call{value: bidderAmount[bidder]}("");
-        require(success, "Couldn't transfer to bidder");
+    /// @dev Requests withdrawal of the staked amount.
+    function requestWithdrawal() external {
```
shaspitz, line 193: To be consistent with [here](https://github.com/primev/mev-commit/blob/d5df1718f53699ed7d50aa6a472e0c9b40277f81/contracts/contracts/validator-registry/ValidatorRegistryV1.sol#L137), this action could be called "unstake". Doesn't matter too much tho, that's just the language I'm used to with other chains

contracts/contracts/ProviderRegistry.sol
```diff
@@ -194,53 +189,37 @@ contract ProviderRegistry is
...
+        require(providerStakes[msg.sender] > 0, "No stake to withdraw");
+        withdrawalRequests[msg.sender] = block.timestamp;
```
shaspitz, line 195: This function shouldn't be callable once it's already been called for the sender 
```suggestion
        require(withdrawalRequests[msg.sender] == 0);
        withdrawalRequests[msg.sender] = block.timestamp;
```

shaspitz (commented):
Nice!

contracts/contracts/ProviderRegistry.sol
```diff
@@ -194,54 +208,57 @@ contract ProviderRegistry is
...
-        providerStakes[provider] = 0;
-        require(stake > 0, "Provider Staked Amount is zero");
-        require(
-            preConfirmationsContract != address(0),
-            "preconf contract not set"
-        );
+    /// @dev Completes the withdrawal of the staked amount.
+    function withdraw() external nonReentrant {
+        require(withdrawalRequests[msg.sender] != 0, "No unstake request");
+        require(block.timestamp >= withdrawalRequests[msg.sender] + withdrawalPeriod, "Period has not passed");
```
shaspitz, line 222: I believe the mev-commit chain's `block.timestamp` is in milliseconds? cc @ckartik 

I'd suggest using blocks for _period of time_ parameters as is done with the validator registry contracts

Mikelle: Good catch with milliseconds! It’s safe to use block.timestamp here. Typically, block.number is recommended when there’s a potential risk of a miner manipulating the timestamp. However, in this case, since the waiting time is 24 hours (not milliseconds), such manipulation is not a concern.

shaspitz: Yes the withdraw period is being set to `24 * 3600`, and the comments specify `withdrawalPeriod` is in seconds, but the `block.timestamp` of our chain is in milliseconds. So the effective withdraw period enforced on this line will be 1000x faster than what's specified. Unless I'm misunderstanding something

Mikelle: You 100% right! I was uploading a fix for that...

shaspitz: The fact that our chain has changed the unit of `block.timestamp` is why I was suggesting to use number of blocks, but timestamp is fine too as long as we specify correct units

contracts/contracts/ProviderRegistry.sol
```diff
@@ -74,16 +81,19 @@ contract ProviderRegistry is
      * @param _feeRecipient The address that receives fee
      * @param _feePercent The fee percentage for protocol
      * @param _owner Owner of the contract, explicitly needed since contract is deployed w/ create2 factory.
+     * @param _withdrawalDelay The withdrawal delay in seconds.
```
shaspitz, line 84:
```suggestion
     * @param _withdrawalDelay The withdrawal delay in milliseconds.
```

shaspitz (approved):
🚀

PR #289, opened 2024-07-30
fix: bugs from testing session

aloknerurkar:
## Describe your changes
Fixes few issues identified in the testing session
- The path for the database directory was not resolved correctly. If user provides `~`, this needs to be resolved to home directory.
- Send SIGINT while killing mev-commit from the launch script. This prevented graceful shutdown of the node. The database is flushed to disk on close, so we should provide the right signal.
- Because the database was not flushed properly in some cases, the internal state was lost. The PR adds fallbacks for depositmanager and autodepositor to check for contract state in the case we dont find things locally. This should provide additional security if the events are not being processed in time.

Fixes # (issue)

- [ ] I have tested this code by deploying the infrastructure and ensuring that commitments are being settled

### DIFF
--- a/p2p/cmd/main.go
+++ b/p2p/cmd/main.go
@@ -374,7 +374,7 @@ func main() {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
-	sigc := make(chan os.Signal)
+	sigc := make(chan os.Signal, 1)
 	signal.Notify(sigc, syscall.SIGINT, syscall.SIGTERM)
 	go func() {
 		<-sigc


Mikelle (approved):
Looks good 👍🏻

launchmevcommit
```diff
@@ -278,4 +295,7 @@ main() {
   wait ${pid}
 }
 
+# Global flag for cleanup
+cleaned_up=false
```
mrekucci, line 299: This is not necessary.

x/util/util.go
```diff
@@ -70,3 +72,20 @@ func NewLogger(lvl, logFmt, tags string, sink io.Writer) (*slog.Logger, error) {
 
 	return logger.With(args...), nil
 }
+
+func ResolveFilePath(path string) (string, error) {
+	if path == "" {
+		return "", fmt.Errorf("path is empty")
+	}
+
+	if strings.HasPrefix(path, "~") {
```
mrekucci, line 81: nit: `strings.HasPrefix(path, "~")` -> `path[0] == '~'`

mrekucci (approved):

PR #293, opened 2024-07-31
feat: automatic protocol fee payout

shaspitz:
## Describe your changes

- Removes prev functionality requiring the protocol fee recipient to submit a tx to receive funds
- Removes ability to not specify a protocol fee recipient 
- Adds automatic fee payout functionality that pays accumulated protocol fees to the appropriate address after a configurable amount of blocks
- Misc documentation/naming refactors


contracts/test/ProviderRegistryTest.sol
```diff
@@ -81,30 +86,36 @@ contract ProviderRegistryTest is Test {
...
         assertEq(providerRegistry.providerRegistered(provider), false);
+        (address recipient, uint256 accumulatedAmount, uint256 lastPayoutBlock, uint256 payoutPeriodBlocks) = bidderRegistry.protocolFeeTracker();
+        assertEq(recipient, feeRecipient);
+        assertEq(payoutPeriodBlocks, protocolFeePayoutPeriodBlocks);
+        assertEq(lastPayoutBlock, block.number);
+        assertEq(accumulatedAmount, 0);
     }
 
-    function testFail_ProviderStakeAndRegisterMinStake() public {
+    function testFailProviderStakeAndRegisterMinStake() public {
```
Mikelle, line 103: What is the purpose of renaming test functions? It seems like Foundry is using the same notation that we have:
https://book.getfoundry.sh/forge/writing-tests

Mikelle: Here is the list from their documentation for the allowed test function names:

https://github.com/ScopeLift/scopelint/blob/1857e3940bfe92ac5a136827374f4b27ff083971/src/check/validators/test_names.rs#L106-L143

shaspitz: These were originally changed because the vs code linter integration was failing to recognize our `.solhintignore` file, and the test files were filled with naming errors. 

imo we should stay consistent to the linter we chose, solhint. Foundry doesn't enforce we use scopelint or any linter at all. Futher on in the first link you've sent, the foundry docs use the naming convention that scopelint would call bad:
```solidity
contract ContractTest is Test {
    uint256 a;
    uint256 b;

    function beforeTestSetup(
        bytes4 testSelector
    ) public pure returns (bytes[] memory beforeTestCalldata) {
        if (testSelector == this.testC.selector) {
            beforeTestCalldata = new bytes[](2);
            beforeTestCalldata[0] = abi.encodePacked(this.testA.selector);
            beforeTestCalldata[1] = abi.encodeWithSignature("setB(uint256)", 1);
        }
    }

    function testA() public {
        require(a == 0);
        a += 1;
    }

    function setB(uint256 value) public {
        b = value;
    }

    function testC() public {
        assertEq(a, 1);
        assertEq(b, 1);
    }
}

```

shaspitz: I was wrong, solhint does have a rule that is consistent with what you're saying: https://protofire.github.io/solhint/docs/rules/naming/foundry-test-functions.html. For this rule tho, looks like we'd need to create a separate `.solhint.json` for the testing folder

Mikelle: Sorry, I was referring to this page, there they are more explicit:
https://book.getfoundry.sh/tutorials/best-practices#tests

shaspitz: Fixed in https://github.com/primev/mev-commit/commit/33d4e1e05a101473820367a1e3e47b6f34678a44

ckartik:
What's going to trigger the fee payouts here?

contracts/contracts/BidderRegistry.sol
```diff
@@ -376,10 +373,10 @@ contract BidderRegistry is
     /**
      * @notice Sets the new fee recipient
      * @dev onlyOwner restriction
-     * @param newFeeRecipient The address to transfer the slashed funds to.
+     * @param newProtocolFeeRecipient The new address to accumulate protocol fees
      */
-    function setNewFeeRecipient(address newFeeRecipient) external onlyOwner {
-        feeRecipient = newFeeRecipient;
+    function setNewProtocolFeeRecipient(address newProtocolFeeRecipient) external onlyOwner {
+        protocolFeeTracker.recipient = newProtocolFeeRecipient;
```
Mikelle, line 379: It's a good practice to generate event in case that function is called
```suggestion
        protocolFeeTracker.recipient = newProtocolFeeRecipient;
        emit ProtocolFeeRecipientUpdated(newProtocolFeeRecipient);
```

contracts/contracts/BidderRegistry.sol
```diff
@@ -391,15 +388,13 @@ contract BidderRegistry is
...
+     * @param newFeePayoutPeriodBlocks The new fee payout period in blocks
      */
-    function withdrawFeeRecipientAmount() external nonReentrant {
-        uint256 amount = feeRecipientAmount;
-        feeRecipientAmount = 0;
-        require(amount != 0, "fee amount is zero");
-        (bool successFee, ) = feeRecipient.call{value: amount}("");
-        require(successFee, "fee transfer failed");
+    function setNewFeePayoutPeriodBlocks(uint256 newFeePayoutPeriodBlocks) external onlyOwner {
+        protocolFeeTracker.payoutPeriodBlocks = newFeePayoutPeriodBlocks;
```
Mikelle, line 397: It's a good practice to generate event in case that function is called
```suggestion
         protocolFeeTracker.payoutPeriodBlocks = newFeePayoutPeriodBlocks;
         emit FeePayoutPeriodBlocksUpdated(newFeePayoutPeriodBlocks);
```

contracts/contracts/ProviderRegistry.sol
```diff
@@ -176,12 +173,12 @@ contract ProviderRegistry is
...
-     * @notice Sets the new fee recipient
+     * @notice Sets a new protocol fee recipient
      * @dev onlyOwner restriction
-     * @param newFeeRecipient The address to transfer the slashed funds to.
+     * @param newFeeRecipient The address of the new protocol fee recipient
      */
-    function setNewFeeRecipient(address newFeeRecipient) external onlyOwner {
-        feeRecipient = newFeeRecipient;
+    function setNewProtocolFeeRecipient(address newFeeRecipient) external onlyOwner {
+        protocolFeeTracker.recipient = newFeeRecipient;
```
Mikelle, line 181: It's a good practice to generate event in case that function is called
```suggestion
        protocolFeeTracker.recipient = newProtocolFeeRecipient;
        emit ProtocolFeeRecipientUpdated(newProtocolFeeRecipient);
```

contracts/contracts/ProviderRegistry.sol
```diff
@@ -201,13 +198,10 @@ contract ProviderRegistry is
...
-     * @dev Reward funds to the fee receipt.
-     */
-    function withdrawFeeRecipientAmount() external nonReentrant {
-        feeRecipientAmount = 0;
-        (bool successFee, ) = feeRecipient.call{value: feeRecipientAmount}("");
-        require(successFee, "fee recipient transfer failed");
+    /// @dev Sets the fee payout period in blocks
+    /// @param _feePayoutPeriodBlocks The new fee payout period in blocks
+    function setFeePayoutPeriodBlocks(uint256 _feePayoutPeriodBlocks) external onlyOwner {
+        protocolFeeTracker.payoutPeriodBlocks = _feePayoutPeriodBlocks;
```
Mikelle, line 204: It's a good practice to generate event in case that function is called
```suggestion
         protocolFeeTracker.payoutPeriodBlocks = _feePayoutPeriodBlocks;
         emit FeePayoutPeriodBlocksUpdated(_feePayoutPeriodBlocks);
```

contracts/contracts/utils/FeePayout.sol
```diff
@@ -0,0 +1,47 @@
...
+        uint256 payoutPeriodBlocks;
+    }
+
+    /// @dev Event emitted when fees are transferred to the recipient.
+    event FeeTransfer(uint256 amount, address recipient);
+
+    /// @dev Initialize a new fee tracker in storage
+    function init(Tracker storage self, address _recipient, uint256 _payoutPeriodBlocks) internal {
+        require(_recipient != address(0), "fee recipient is zero");
+        require(_payoutPeriodBlocks > 0, "pay period must be positive");
```
Mikelle, line 23:
```suggestion
        require(_payoutPeriodBlocks != 0, "pay period must be positive");
```

Mikelle (commented):
What if the oracle didn’t perform any slashes or rewards, but we still want to withdraw protocol revenue? Do we want to have a mechanism to withdraw those funds without relying on the slashing/rewarding mechanism?

shaspitz:
> What if the oracle didn’t perform any slashes or rewards, but we still want to withdraw protocol revenue? Do we want to have a mechanism to withdraw those funds without relying on the slashing/rewarding mechanism?

This edge case most likely won't happen as after 1 hour, it would take a single slash/reward invocation to trigger the fee payout. But it wouldn't hurt to cover the edge case anyways 👍, see https://github.com/primev/mev-commit/commit/9bc69763d5a5aa31702e133f160ee5aa3af31ab6

contracts/contracts/utils/FeePayout.sol
```diff
@@ -0,0 +1,47 @@
...
+        /// @dev Accumulated fees since last payout
+        uint256 accumulatedAmount;
+        /// @dev Block number when the last fee payout was made
+        uint256 lastPayoutBlock;
+        /// @dev Min number of blocks between payouts
+        uint256 payoutPeriodBlocks;
+    }
+
+    /// @dev Event emitted when fees are transferred to the recipient.
+    event FeeTransfer(uint256 amount, address recipient);
```
Mikelle, line 18: I think it will be goot to indexing recipient, so we can filter events by specific recipient addresses when search through logs, which can be very useful for tracking and auditing purposes.

Mikelle (approved):

PR #295, opened 2024-07-31
Fixes bug with treasury getter of slashed funds and sets fee to 5% for slashing

ckartik:
## Describe your changes
- Previously, we were setting the Fee allocated to the treasury due to slashing events to zero, before pulling the funds. We update this to ensure we keep a temporary variable that stores the value meant for the treasury during retrieval. We also set the slashing percentage reward for the protocol to 5%.

- [ ] I have tested this code by deploying the infrastructure and ensuring that commitments are being settled

### DIFF
--- a/contracts/scripts/DeployScripts.s.sol
+++ b/contracts/scripts/DeployScripts.s.sol
@@ -26,6 +26,7 @@ contract DeployScript is Script {
             0x68bC10674b265f266b4b1F079Fa06eF4045c3ab9
         );
         uint16 feePercent = 2;
+        uint16 slashPercent = 5;
         uint64 commitmentDispatchWindow = 2000;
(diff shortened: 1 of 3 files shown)


contracts/scripts/DeployScripts.s.sol
```diff
@@ -26,6 +26,7 @@ contract DeployScript is Script {
             0x68bC10674b265f266b4b1F079Fa06eF4045c3ab9
         );
         uint16 feePercent = 2;
+        uint16 slashPercent = 5;
```
shaspitz, line 29: The above change will be n/a with https://github.com/primev/mev-commit/pull/293, but we can use this change

ckartik:
Duplicate given shawns other changes.

Issue #309, opened 2024-08-02
Add validator API


Mikelle:
* Get current epoch
* Get info is mev-commit validator is participating in next epoch or not

PR #310, opened 2024-08-02
feat: add validator api

Mikelle:
## Describe your changes

Add validator API

Fixes https://github.com/primev/mev-commit/issues/309


p2p/cmd/main.go
```diff
@@ -238,6 +238,18 @@ var (
 		},
 	})
 
+	optionValidatorRouterAddr = altsrc.NewStringFlag(&cli.StringFlag{
+		Name:    "validator-router-contract",
+		Usage:   "address of the validator router contract",
+		EnvVars: []string{"MEV_COMMIT_VALIDATOR_ROUTER_ADDR"},
+		Action: func(ctx *cli.Context, s string) error {
```
shaspitz, line 245: Should we supply a default value for deployment on Holesky? If it exists

p2p/gen/openapi/validatorapi/v1/validatorapi.swagger.yaml
```diff
@@ -0,0 +1,102 @@
...
+          description: A successful response.
+          schema:
+            $ref: '#/definitions/v1EpochResponse'
+        default:
+          description: An unexpected error response.
+          schema:
+            $ref: '#/definitions/googlerpcStatus'
+  /v1/validator/get_validators:
+    get:
+      summary: GetValidators
```
shaspitz, line 29: Is there a reason why we separate this out from `GetEpoch`? Ie. why not just have a single endpoint that returns the validators of the latest epoch? 

Reasoning here should prob be in our docs

p2p/pkg/rpc/validator/service.go
```diff
@@ -0,0 +1,192 @@
...
+	finalizedEpoch, err := strconv.ParseUint(checkpointsResp.Data.Finalized.Epoch, 10, 64)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "parsing finalized epoch: %v", err)
+	}
+
+	s.logger.Info("fetched epoch data", "current_epoch", currentEpoch, "current_justified_epoch", currentEpoch, "previous_justified_epoch", previousJustifiedEpoch, "finalized_epoch", finalizedEpoch)
+	s.metrics.FetchedEpochDataCount.Inc()
+
+	return &validatorapiv1.EpochResponse{
+		CurrentEpoch:           currentEpoch + 1,
```
shaspitz, line 110: Do we need to expose all three of these values to the user? Is there some tradeoffs we should communicate to them?

p2p/pkg/rpc/validator/service.go
```diff
@@ -0,0 +1,192 @@
...
+		return nil, status.Errorf(codes.Internal, "unexpected status code: %v", resp.StatusCode)
+	}
+
+	var checkpointsResp FinalityCheckpointsResponse
+	if err := json.NewDecoder(resp.Body).Decode(&checkpointsResp); err != nil {
+		s.logger.Error("decoding response", "error", err)
+		return nil, status.Errorf(codes.Internal, "decoding response: %v", err)
+	}
+
+	currentEpoch, err := strconv.ParseUint(checkpointsResp.Data.CurrentJustified.Epoch, 10, 64)
```
shaspitz, line 91: To avoid naming confusion

```suggestion
	currentjustifiedEpoch, err := strconv.ParseUint(checkpointsResp.Data.CurrentJustified.Epoch, 10, 64)
	currentEpoch := currentJustifiedEpoch + 1
```

p2p/pkg/rpc/validator/service.go
```diff
@@ -0,0 +1,192 @@
...
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "parsing previous justified epoch: %v", err)
+	}
+
+	finalizedEpoch, err := strconv.ParseUint(checkpointsResp.Data.Finalized.Epoch, 10, 64)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "parsing finalized epoch: %v", err)
+	}
+
+	s.logger.Info("fetched epoch data", "current_epoch", currentEpoch, "current_justified_epoch", currentEpoch, "previous_justified_epoch", previousJustifiedEpoch, "finalized_epoch", finalizedEpoch)
```
shaspitz, line 106: `current_epoch` here should be different than `current_justified_epoch` right?

p2p/rpc/validatorapi/v1/validatorapi.proto
```diff
@@ -0,0 +1,68 @@
...
+            title: "GetValidatorsRequest";
+            description: "Request for GetValidators";
+        }
+    };
+    uint64 epoch = 1;
+}
+
+message SlotInfo {
+    string BLSKey = 1;
+    bool isActive = 2;
```
shaspitz, line 63: IsActive already has meaning in the context of L1 consensus so something like "is opted in" or "is mev-commit opted in" may be better here
```suggestion
    bool isOptedIn = 2;
```

p2p/pkg/rpc/validator/service.go
```diff
@@ -0,0 +1,192 @@
...
+	}
+	for i, duty := range dutiesResp.Data {
+		slot, err := strconv.ParseUint(duty.Slot, 10, 64)
+		if err != nil {
+			s.logger.Error("parsing slot number", "error", err)
+			continue
+		}
+		validators[slot] = &validatorapiv1.SlotInfo{
+			BLSKey:   duty.Pubkey,
+			IsActive: areValidatorsOptedIn[i],
```
shaspitz, line 182: see below comment on naming of `IsActive`

shaspitz (commented):
Did querying the router contract end up working? ie. this is working e2e?

shaspitz:
We'll want to add some docs on how bidders can utilize these new endpoints to have better UX, along with any timing considerations

p2p/cmd/main.go
Mikelle: We don't do it for oracle, so I didn't add default variable here (as it also comes with private key for the alchemy)

p2p/gen/openapi/validatorapi/v1/validatorapi.swagger.yaml
Mikelle: Asked Evan, he said that it's sufficient to `GetEpoch` be done automatically in `GetValidators` request

Mikelle:
Thank you for the review! I definitely agree on the necessity of documentation. I wanted to add it once we have the latest smart contract set and that branch finalized and merged.

Mikelle:
> Did querying the router contract end up working? ie. this is working e2e?
Router is working with fake ValidatorRegistry (which is always returns true)

shaspitz (approved):
Looks great!

PR #332, opened 2024-08-12
feat: delegation for provider registration/stake

shaspitz:
## Describe your changes

Accounts can now register and stake on behalf of another provider account, enabling provider registration migration


contracts/contracts/ProviderRegistry.sol
```diff
@@ -288,6 +303,23 @@ contract ProviderRegistry is
         require(withdrawalRequests[provider] == 0, "Pending withdrawal request");
     }
 
+    function _stake(address provider) internal {
+        require(providerRegistered[provider], "Provider not registered");
```
Mikelle, line 307: Do we need that check? If the provider has deposited some funds, they will not be able to deposit more because we don’t reset this value to false once the provider withdraws their funds.

shaspitz: Having this check is consistent with [current functionality in main](https://github.com/primev/mev-commit/blob/cb1015776575841f8bdbc5cf94714db5ac27e62b/contracts/contracts/ProviderRegistry.sol#L141). Sounds like you're describing a bug that's unrelated to the change in this PR, correct?

Mikelle: Right

shaspitz: Could you elaborate what the bug is in a separate issue? The issue isn't immediately clear to me, but I believe the original intended UX flow is that a bidder would call `registerAndStake` first (setting `providerRegistered` mapping to true), then subsequently call `stake` whenever funds need to be added

Mikelle (approved):

PR #343, opened 2024-08-15
feat: introduce OTEL tracing

mrekucci:
## Describe your changes

Introduce Open Telemetry tracing.


mrekucci:
> LGTM! We should also instrument the p2p request-response messages. There is a way to pass tracing information using headers.

This can come up in the later stages, I'd strongly advocate being careful what is instrumented with traces as it can easily cause a mess.

PR #372, opened 2024-08-28
chore: gas adjustments

aloknerurkar:
## Describe your changes
The failures on the integration tests are happening because of the gas we set on the transactions. Some of these transactions need more gas.

Fixes # (issue)

### DIFF
--- a/testing/pkg/tests/preconf/preconf.go
+++ b/testing/pkg/tests/preconf/preconf.go
@@ -423,8 +423,15 @@ func getRandomBid(
 		// skip
 	default:
 		start := rand.Intn(txCount)
-		end := rand.Intn(min(start+5, txCount-1)-start) + start + 1 // Limit to max of 4 transactions.
-		transactions = transactions[start:end]
+		// we select a random number of transactions to bundle starting from the start index
+		// in that order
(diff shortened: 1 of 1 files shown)


testing/pkg/tests/preconf/preconf.go
```diff
@@ -412,22 +413,29 @@ func getRandomBid(
 		store.Insert(blkKey(blkNum), blk)
 	}
 
-	if len(blk.(*types.Block).Transactions()) == 0 {
+	transactions := blk.(*types.Block).Transactions()
+	txCount := len(transactions)
+
+	if txCount == 0 {
 		return nil, errNoTxnsInBlock
 	}
```
mrekucci, line 422: I'd recommend the following simplification:
```Golang
t := transactions
switch n := len(t); n {
case 0:
	return nil, errNoTxnsInBlock
case 1:
	// skip...
default:
	s := rand.Intn(n)
	e := rand.Intn(min(s+4, n-1)-s) + s + 1 // Limit to max of 4 transactions.
	t := t[s:e]
}

for _, txn := range t {
	// ...
```

mrekucci (approved):

PR #384, opened 2024-09-05
fix: retry on failed event subscription

aloknerurkar:
## Describe your changes
Retry subscription to logs on error. Allow sufficient time before we declare error.

Fixes # (issue)

### DIFF
--- a/x/contracts/events/publisher/ws.go
+++ b/x/contracts/events/publisher/ws.go
@@ -4,10 +4,15 @@ import (
 	"context"
 	"log/slog"
 	"math/big"
+	"time"
 
 	"github.com/ethereum/go-ethereum"
 	"github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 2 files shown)


kant777 (approved):

mrekucci (approved):
I would suggest considering this package to simplify the solution and make it more robust: https://github.com/cenkalti/backoff

mrekucci:
> > I would suggest considering this package to simplify the solution and make it more robust: https://github.com/cenkalti/backoff
> 
> I am not really sure if we need exponential backoff here. This is currently problem with connection to our settlement layer. We dont really have any rate-limiting on geth. So the only reason I see this dropping is because of connection drop on client side. So not sure if exponential backoff will help. Wdyt?

You don't have to opt for exponential repetition, the package is flexible enough so that it can be set to fixed periodic repetition.

PR #410, opened 2024-09-24
refactor: custom errors for core contracts

### LINKED ISSUE #376: Migrate solidity to use custom errors
https://github.com/primev/mev-commit/pull/375 will introduce solidity custom error unmarshaling for our geth library client code. Following that PR being merged we should replace all instances of solidity `require` with custom errors. This should be a largely trivial change for core contracts but might require some additional client changes for L1 (validator registry) contracts.

shaspitz:

Closes #376 


contracts/contracts/core/BidderRegistry.sol
```diff
@@ -26,10 +26,7 @@ contract BidderRegistry is
      * @dev Modifier to restrict a function to only be callable by the preconfManager contract.
      */
     modifier onlyPreconfManager() {
-        require(
-            msg.sender == preconfManager,
-            "sender is not preconf contract"
-        );
+        require(msg.sender == preconfManager, SenderIsNotPreconfManager());
```
aloknerurkar, line 29: I think it would be better if we add some arguments here. For eg. the sender address. The goal here is to have more information, but currently this is not giving any new information.

contracts/contracts/core/BidderRegistry.sol
```diff
@@ -143,10 +139,7 @@ contract BidderRegistry is
         uint256 len = windows.length;
         for (uint256 i = 0; i < len; ++i) {
             uint256 window = windows[i];
-            require(
-                window < currentWindow,
-                "withdraw after window settled"
-            );
+            require(window < currentWindow, WithdrawAfterWindowSettled());
```
aloknerurkar, line 142: Again, here I think we should print both the windows. If we dont know the values, we wont be able to debug what is going wrong here.

contracts/contracts/core/BidderRegistry.sol
```diff
@@ -161,7 +154,7 @@ contract BidderRegistry is
         }
 
         (bool success, ) = msg.sender.call{value: totalAmount}("");
-        require(success, "transfer to bidder failed");
+        require(success, TransferToBidderFailed());
```
aloknerurkar, line 157: Again, I think the amount and bidder address both need to be added here

contracts/contracts/core/BidderRegistry.sol
```diff
@@ -179,10 +172,8 @@ contract BidderRegistry is
         uint256 residualBidPercentAfterDecay
     ) external nonReentrant onlyPreconfManager whenNotPaused {
         BidState storage bidState = bidPayment[commitmentDigest];
-        require(
-            bidState.state == State.PreConfirmed,
-            "bid not preconfirmed"
-        );
+        require(bidState.state == State.PreConfirmed, BidNotPreConfirmed());
```
aloknerurkar, line 175: BidID should be added here.

contracts/contracts/core/BidderRegistry.sol
```diff
@@ -228,16 +219,14 @@ contract BidderRegistry is
         bytes32 bidID
     ) external nonReentrant onlyPreconfManager whenNotPaused {
         BidState storage bidState = bidPayment[bidID];
-        require(
-            bidState.state == State.PreConfirmed,
-            "The bid was not preconfirmed"
-        );
+        require(bidState.state == State.PreConfirmed, BidWasNotPreConfirmed());
```
aloknerurkar, line 222: Again bidID should be added here.

contracts/contracts/core/ProviderRegistry.sol
```diff
@@ -267,20 +268,20 @@ contract ProviderRegistry is
 
     /// @dev Ensure the provider's balance is greater than minStake and no pending withdrawal
     function isProviderValid(address provider) public view {
-        require(providerStakes[provider] >= minStake, "Insufficient stake");
-        require(withdrawalRequests[provider] == 0, "Pending withdrawal request");
+        require(providerStakes[provider] >= minStake, InsufficientStake());
```
aloknerurkar, line 271: I think this is a general comment for all of these errors. I think we should add more arguments here that will help identify the problems. For example the provider address here.

contracts/contracts/core/BidderRegistry.sol
shaspitz: I was trying to get past this PR quickly and minimally introduce custom errors just for gas savings, but you bring up a valid point 👍 I've added args to most errors in this PR

aloknerurkar (approved):
Nice!

Mikelle (approved):

PR #416, opened 2024-09-27
perf: gas optimizations for vanillaRegistry and mevCommitAVS

shaspitz:
## Describe your changes

Introduces local storage variables to `VanillaRegistry` and `MevCommitAVS` functions to reduce storage reads and therefore gas usage. A few existing variables were changed from `memory` to storage to also save on gas. 

This PR should be entirely a refactor so tests should remain unchanged.


contracts/contracts/validator-registry/avs/MevCommitAVS.sol
```diff
@@ -517,9 +522,10 @@ contract MevCommitAVS is IMevCommitAVS, MevCommitAVSStorage,
```
shaspitz, line 522: `_freeze` above is handled in https://github.com/primev/mev-commit/pull/415

mrekucci (approved):

PR #426, opened 2024-10-03
feat: introduce the --nat-ip flag for launchmevcommit

mrekucci:
## Describe your changes

Introduces the `--nat-ip` flag for launchmevcommit to allow providers to set ip and/or port if they are behind NAT.


aloknerurkar (approved):
Nice!

PR #434, opened 2024-10-15
chore: bridge deploy scripts

shaspitz:
## Describe your changes

* Adds up-to-date deploy scripts which should be used for the bridge contracts
* These scripts follow a more organized plan around genesis allocations in notion (ie. which accounts get allocated/sent ether during contract setup). 
* Deletes outdated scripts


bridge/standard/bridge-v1/deploy_contracts.sh
```diff
@@ -1,139 +0,0 @@
-#!/bin/sh
```
shaspitz, line 1: This file is outdated, we should now do this config in nomad/solidity

contracts/entrypoint.sh
```diff
@@ -37,61 +37,4 @@ if [ "${DEPLOY_TYPE}" = "core" ]; then
         --force \
         --json \
         --via-ir
-
```
shaspitz, line 40: Deleting these cases since the nomad file can just call forge commands inline and reduce complexity

contracts/scripts/standard-bridge/DeployStandardBridge.s.sol
```diff
@@ -13,75 +13,145 @@ import {Strings} from "@openzeppelin/contracts/utils/Strings.sol";
...
-        Allocator allocator = Allocator(payable(allocatorProxy));
-        console.log("Allocator deployed to:", address(allocator));
+            address(settlementGateway));
 
-        allocator.addToWhitelist(address(gateway));
-        console.log("Settlement gateway has been whitelisted. Gateway contract address:", address(gateway));
+        allocator.addToWhitelist(address(settlementGateway));
+        console.log("Settlement gateway has been whitelisted");
 
         string memory jsonOutput = string.concat(
```
mrekucci, line 92: nit: maye consider using something like the following: https://github.com/bmeredith/solidity-json-writer

mrekucci (approved):

contracts/scripts/standard-bridge/DeployStandardBridge.s.sol
shaspitz: Nice this makes sense, but we'd seemingly have to make it compatible with `forge-std/console.sol` output. That repo is hardhat based

PR #441, opened 2024-10-17
feat: deploy bridge contracts

mrekucci:


aloknerurkar (approved):

infrastructure/nomad/playbooks/deploy.yml
```diff
@@ -495,7 +495,7 @@
             echo "${ALLOC}" | jq --arg address "0x${ADDRESS}" '
```
shaspitz, line 495: We should change this step to not use a for-loop. An essential property of the mev-commit chain is that only a single account will receive an initial ETH allocation, the contract deployer, which will later fund the allocator contract.

infrastructure/nomad/playbooks/deploy.yml
```diff
@@ -495,7 +495,7 @@
...
               . + {
                 ($address): {
-                    "balance": "0x10000000000000000000000000000"
+                    "balance": "0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFD4A51000FDA0FFFF"
```
shaspitz, line 498: Does this match `DEPLOYER_GENESIS_ALLOCATION` from `contracts/scripts/standard-bridge/DeployStandardBridge.s.sol`?

infrastructure/nomad/playbooks/templates/jobs/contracts-deployer.nomad.j2
```diff
@@ -92,6 +97,8 @@ job "{{ job.name }}" {
 
       template {
         data = <<-EOH
+          L1_FINALIZATION_FEE="1"
```
shaspitz, line 100: These values are all good for devnet, but for holesky and mainnet deployment these values will need to be higher (on average, over time) than what the relayer pays to call `finalizeTransfer`

infrastructure/nomad/playbooks/deploy.yml
mrekucci: Yes, otherwise it would fail when contracts are deployed.

infrastructure/nomad/playbooks/templates/jobs/contracts-deployer.nomad.j2
mrekucci: What amount would be sufficient for the mainnet?

shaspitz: We'll need to test or calculate what the relayer pays in ETH to call finalizeTransfer on average, over time. Lets call that `avgCost` for each chain. FinalizationFee must be set to 2*`avgCost` or some margin of safety above the cost that the relayer will incur on average for each tx

PR #443, opened 2024-10-18
fix: increase bid processing timeout

aloknerurkar:
## Describe your changes
The timeout for `ProcessBid` operation which communicates with the provider API is too small. Increasing it to 15 secs.

Fixes # (issue)

### DIFF
--- a/p2p/pkg/preconfirmation/preconfirmation.go
+++ b/p2p/pkg/preconfirmation/preconfirmation.go
@@ -293,7 +293,7 @@ func (p *Preconfirmation) handleBid(
 	}()
 
 	// try to enqueue for 5 seconds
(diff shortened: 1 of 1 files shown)


ckartik (approved):
We might want to keep this the length of two slots, so 30 seconds. Up to you, will approve this as is. Maybe we should allow this to be passed in by the provider as a parameter though?

p2p/pkg/preconfirmation/preconfirmation.go
```diff
@@ -293,7 +293,7 @@ func (p *Preconfirmation) handleBid(
 	}()
 
 	// try to enqueue for 5 seconds
```
ckartik, line 295:
```suggestion
	// try to enqueue for 15 seconds
```

PR #444, opened 2024-10-18
feat: redis consensus

Mikelle:
## Describe your changes

Implement redis geth consensus integration


cl/redisapp/steps.go
```diff
@@ -0,0 +1,386 @@
...
+
+type StepsManager struct {
+	stateManager StateManager
+	engineCl     EngineClient
+	logger       Logger
+	buildDelay   time.Duration
+	ctx          context.Context
+}
+
+func NewStepsManager(ctx context.Context, stateManager StateManager, engineCl EngineClient, logger Logger, buildDelay time.Duration) *StepsManager {
```
mrekucci, line 29: Same as above, this constructor is unnecessary.

mrekucci (commented):
The CI is failing: https://github.com/primev/mev-commit/actions/runs/11412548053/job/31758694912?pr=444

cl/redisapp/follower.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+func (f *Follower) followerLoop(ctx context.Context) {
+	err := f.stateManager.CreateConsumerGroup(ctx)
+	if err != nil {
+		f.logger.Error("Failed to create consumer group", "error", err)
+		return
+	}
+
+	for {
```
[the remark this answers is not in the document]
Mikelle: Does it make sense to check if I already have a checker for the child context? As far as I know, if the parent context is canceled, the child context will be canceled automatically.

cl/redisapp/rapp.go
```diff
@@ -0,0 +1,178 @@
...
+		TTL:   100 * time.Millisecond,
+		Wait:  200 * time.Millisecond,
+		Key:   "rapp_leader_election",
+	}
+
+	procLeader, promotedCh, demotedCh, erroredCh := leader.NewLeader(leaderOpts)
+
+	stateManager := NewRedisStateManager(instanceID, redisClient, logger, genesisBlockHash)
+
+	var wg sync.WaitGroup
```
aloknerurkar, line 93: Sharing this waitgroup across multiple objects seems to be inviting trouble. Not sure why this structure is needed. The leader and follower routines dont need to be initialized here. Instead the leader election should start these routines based on the events. They can return separate channels to signal that they have stopped and can use separate contexts/channels to stop them.

cl/redisapp/rapp.go
```diff
@@ -0,0 +1,178 @@
...
+		InstanceID:     instanceID,
+		wg:             &wg,
+		stateManager:   stateManager,
+		stepsManager:   stepsManager,
+		leaderElection: procLeader,
+		logger:         logger,
+	}
+
+	// Initialize LeaderElectionHandler
+	leaderElectionHandler := NewLeaderElectionHandler(
```
aloknerurkar, line 124: Also, not sure why all these constructors are exported. They are only used within this package. Ideally they should reside in an internal package inside the redisapp as they are not really useful on their own. Everything is dependent on redis.

Better modularization would be to not have these packages depend on redis, in which case they could be exported and then the storage/communication interface is defined in such a way that we can use different backends.

cl/redisapp/retry.go
```diff
@@ -0,0 +1,113 @@
...
+func backoff(attempt int) time.Duration {
+	backoff := float64(initialBackoff) * math.Pow(2, float64(attempt))
+	if backoff > float64(maxBackoff) {
+		backoff = float64(maxBackoff)
+	}
+	return time.Duration(backoff)
+}
+
+func retryWithInfiniteBackoff(ctx context.Context, operation func() (bool, error)) (bool, error) {
+	retryMutex.Lock()
```
aloknerurkar, line 32: If the redis client doesnt provide synchronization, this external mutex doesnt really help in my opinion. If it does not, then there should be synchronization at the package level where it is used.

cl/redisapp/rapp.go
```diff
@@ -0,0 +1,178 @@
...
+
+	// Initialize leader election
+	leaderOpts := leader.Opts{
+		Redis: redisClient,
+		TTL:   100 * time.Millisecond,
+		Wait:  200 * time.Millisecond,
+		Key:   "rapp_leader_election",
+	}
+
+	procLeader, promotedCh, demotedCh, erroredCh := leader.NewLeader(leaderOpts)
```
aloknerurkar, line 89: This could be initialized inside the leaderelection. That way we dont have to pass all these as arguments.

aloknerurkar (commented):
Still not done.

cl/redisapp/retry.go
```diff
@@ -0,0 +1,113 @@
...
+	maxBackoff     = 30 * time.Second       // Maximum backoff duration
+)
+
+var (
+	retryMutex              sync.Mutex
+	ErrFailedAfterNAttempts = errors.New("operation failed after N attempts")
+)
+
+// Backoff function implementing an exponential backoff strategy
+func backoff(attempt int) time.Duration {
```
mrekucci, line 23: nit: there is nice package that implements exponential backoff that can simplify the code: https://github.com/cenkalti/backoff

cl/redisapp/retry.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+func retryWithInfiniteBackoff(ctx context.Context, operation func() (bool, error)) (bool, error) {
+	retryMutex.Lock()
+	defer retryMutex.Unlock()
+
+	for attempt := 0; ; attempt++ {
+		select {
+		case <-ctx.Done():
+			log.Println("Context canceled, stopping retries.")
```
mrekucci, line 38: nit: I'd recommend using `log/slog` instead; the same is below.

cl/redisapp/state.go
```diff
@@ -0,0 +1,281 @@
...
+	s.blockStateMutex.Lock()
+	defer s.blockStateMutex.Unlock()
+
+	// Return a copy of the state to prevent external modification
+	return *s.blockBuildState
+}
+
+func (s *RedisStateManager) CreateConsumerGroup(ctx context.Context) error {
+	if err := s.redisClient.XGroupCreateMkStream(ctx, redisStreamName, s.groupName, "0").Err(); err != nil {
+		if !strings.Contains(err.Error(), "BUSYGROUP") {
```
mrekucci, line 226: Doesn't the redis client export the errors that can be checked, instead of checking part of the error string?

cl/redisapp/state.go
```diff
@@ -0,0 +1,281 @@
...
+	args := &redis.XReadGroupArgs{
+		Group:    s.groupName,
+		Consumer: s.consumerName,
+		Streams:  []string{redisStreamName, string(msgType)},
+		Count:    1,
+		Block:    time.Second,
+	}
+
+	messages, err := s.redisClient.XReadGroup(ctx, args).Result()
+	if err != nil && err != redis.Nil {
```
mrekucci, line 263: Use `errors.Is` instead.

cl/redisapp/state.go
Mikelle: For this case no:

https://github.com/redis/go-redis/blob/1ed936eb097d10f3aaf40a44d282778f879eac32/commands_test.go#L6032

Mikelle: https://github.com/redis/go-redis/issues/2265

cl/redisapp/rapp.go
Mikelle: I improved the package structure by refining the splitting and eliminating unnecessary functions and variables exports. I agree that refactoring to decouple the packages from Redis and defining a storage/communication interface for different backends would enhance modularity, but not sure, is it really needed for that PR (as we are not moving away from redis for that in next 1-2 months) and maybe it could be done later?

aloknerurkar (commented):
I find it hard to reason about the leaderelection pkg. I think it can be simplified. There are few issues I think with having 2 separate goroutines especially when they can be started/stopped multiple times and communicating with the use of this sync channel. This can be avoided by having a single worker.

```
package leaderelection

import (
	"context"
	"sync/atomic"
	"time"

	"github.com/ethereum/go-ethereum/log"
	"github.com/primev/mev-commit/cl/redisapp/state"
	"github.com/primev/mev-commit/cl/redisapp/types"
	"github.com/redis/go-redis/v9"
)

type worker struct {
	isLeader     atomic.Bool
	promotedTime time.Time
	syncCheck    time.Time
	stateManager state.StateManager
	logger       *slog.Logger
}

func New(stateManager state.StateManager, logger *slog.Logger) (*worker, error) {
	if err := stateManager.CreateConsumerGroup(context.Background()); err != nil {
		return nil, err
	}
	return &worker{
		stateManager: stateManager,
		logger:       logger,
	}
}

func (s *worker) selectionWorker(ctx context.Context, redisClient *redis.Client) {
	electionOpts := leader.Opts{
		Redis: redisClient,
		TTL:   100 * time.Millisecond,
		Wait:  200 * time.Millisecond,
		Key:   "rapp_leader_election",
	}

	proc, promotedC, demotedC, errC := leader.New(electionOpts)
	defer proc.Stop()

	for {
		select {
		case <-ctx.Done():
			return
		case p := <-promotedC:
			s.promotedTime = p
			s.isLeader.Store(true)
		case <-demotedC:
			s.isLeader.Store(false)
		case err := <-errC:
			log.Error("leader election error", "error", err)
		}
	}
}

func (s *worker) doWork(ctx context.Context) {
	for {
		select {
		case <-ctx.Done():
			return
		default:
		}

		if s.isLeader.Load() {
			if err := s.leaderWork(ctx); err != nil {
				s.logger.Error("failed to do leader work", "error", err)
			}
		} else {
			if err := s.followerWork(ctx); err != nil {
				s.logger.Error("failed to do follower work", "error", err)
			}
		}
	}
}

func (w *worker) leaderWork(ctx context.Context) error {
	if w.syncCheck.IsZero() || w.promotedTime.After(w.syncCheck) {
		if err := w.followerWork(ctx); err != nil {
			w.logger.Error("failed to do follower work", "error", err)
			return err
		}
	}

	w.syncCheck = time.Now()

	// Do leader work sequentially
	// 1. Get state
	// 2. Build block
	// 3. Finalize block
}

func (w *worker) followerWork(ctx context.Context) error {
	for {
		// Read all messages pending and new
		messages, err := w.stateManager.ReadMessagesFromStream(ctx, types.RedisMsgTypePending)
		if err != nil {
			w.logger.Error("failed to read pending messages from stream", "error", err)
			continue
		}

		if len(messages) == 0 || len(messages[0].Messages) == 0 {
			messages, err = w.stateManager.ReadMessagesFromStream(ctx, types.RedisMsgTypeNew)
			if err != nil {
				w.logger.Error("failed to read new messages from stream", "error", err)
				continue
			}
		}

		if len(messages) == 0 || len(messages[0].Messages) == 0 {
			return nil
		}

		// Process messages
		for _, msg := range messages {
			// Process message
		}
}
```

cl/redisapp/leaderfollower/leaderfollower.go
```diff
@@ -0,0 +1,323 @@
...
+		if err != nil {
+			lfm.logger.Error("Follower: Error reading new messages", "error", err)
+			return nil, err
+		}
+	}
+
+	return messages, nil
+}
+
+func (lfm *LeaderFollowerManager) Stop() error {
```
aloknerurkar, line 306: Stop function doesnt actually stop anything. This is confusing. If user thinks he can call Stop and stop this, he would be blocked forever.

Secondly, it is better practice to do something like:
```
closed := make(chan struct{})
go func() {
  defer close(closed)
  wg.Wait()
}()

select {
  case <-time.After(5*time.Second):
    logger.Error("")
    return errors.New("workers still running")
  case <-closed:
    return nil
}
```

This way the wait is best effort and closing of the calling process is not blocked.

cl/redisapp/leaderfollower/leaderfollower.go
```diff
@@ -0,0 +1,323 @@
...
+	for {
+		select {
+		case <-ctx.Done():
+			lfm.logger.Debug("Leader: Exiting")
+			return nil
+		default:
+			bbState := lfm.stateManager.GetBlockBuildState(ctx)
+			currentStep := bbState.CurrentStep
+
+			switch currentStep {
```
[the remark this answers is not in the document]
Mikelle: This logic ensures that the leader restarts at StepFinalizeBlock if it fails during that step. For example, if only one leader is active at a time and fails at this step, we want the restart to begin from that point rather than from the beginning, as the block was already sent to the followers, and they will attempt to apply it to their Geth instance

cl/redisapp/leaderfollower/leaderfollower.go
```diff
@@ -0,0 +1,323 @@
...
+				if err := lfm.stateManager.ResetBlockState(ctx); err != nil {
+					lfm.logger.Error("Leader: Failed to reset block state", "error", err)
+					return err
+				}
+			}
+		}
+	}
+}
+
+func (lfm *LeaderFollowerManager) ensureStateSynchronized(ctx context.Context) (bool, error) {
```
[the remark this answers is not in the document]
Mikelle: In your case, there could be a situation where the leader node is out of sync. The entire network will stop generating blocks because all nodes will be waiting for the leader to finally sync.

cl/redisapp/state/state.go
```diff
@@ -0,0 +1,281 @@
...
+}
+
+type StateManager interface {
+	SaveExecutionHead(ctx context.Context, head *types.ExecutionHead) error
+	LoadExecutionHead(ctx context.Context) (*types.ExecutionHead, error)
+	LoadOrInitializeBlockState(ctx context.Context) error
+	SaveBlockState(ctx context.Context) error
+	ResetBlockState(ctx context.Context) error
+	SaveExecutionHeadAndAck(ctx context.Context, head *types.ExecutionHead, messageID string) error
+	SaveBlockStateAndPublishToStream(ctx context.Context, bsState *types.BlockBuildState) error
```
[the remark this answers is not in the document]
Mikelle: It is used here:

https://github.com/primev/mev-commit/blob/feat/redis-consensus/cl/redisapp/blockbuilder/blockbuilder.go#L196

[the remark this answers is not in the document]
Mikelle: RecoverLeaderState could be deleted, it is not used

cl/redisapp/leaderfollower/leaderfollower.go
```diff
@@ -0,0 +1,314 @@
...
+	for {
+		select {
+		case <-ctx.Done():
+			lfm.logger.Debug("Leader: Exiting")
+			return nil
+		default:
+			bbState := lfm.stateManager.GetBlockBuildState(ctx)
+			currentStep := bbState.CurrentStep
+
+			switch currentStep {
```
aloknerurkar, line 173: nit: the redundant calls to ResetBlockState can be removed by having an anonymous function around the switch case:

```
err := func() error {
  switch...
}()
if err != nil {
  err = errors.Join(err, lfm.ResetBlockState())
}
```
For the default cast you can return a new error so that this is executed.

aloknerurkar (approved):
Approving this for now. But please add unit tests for the code that is not covered asap

cl/redisapp/leaderfollower/leaderfollower.go
Mikelle: I still have to keep those calls, as when error is happening during `StepFinalizeBlock`, I don't want to reset a state.

PR #455, opened 2024-10-29
fix: wait for finalization

aloknerurkar:
## Describe your changes
Override the ethclient to look for finalized blocks only to verify state of operations.

Fixes # (issue)


x/contracts/ethwrapper/client.go
```diff
@@ -0,0 +1,235 @@
...
+		cli EthClient
+	}
+
+	// Options.
+	blockNumberDrift int
+	winnersOverride  []string
+	maxRetries       int
+}
+
+func (c *Client) RawClient() *ethclient.Client {
```
mrekucci, line 101: nit: missing method documentation.

mrekucci (approved):

bridge/standard/pkg/node/node.go
```diff
@@ -243,9 +243,18 @@ func (n *Node) createGatewayContract(
 		return fmt.Errorf("failed to parse contract ABI: %w", err)
 	}
 
+	wrappedClient, err := ethwrapper.NewClient(
+		logger.With("component", fmt.Sprintf("%s/ethwrapper", component)),
+		[]string{rpcURL},
+		ethwrapper.EthClientWithBlockNumberDrift(2*32), // 2 epochs
```
shaspitz, line 249: By using `ethwrapper.EthClientWithBlockNumberDrift`, we're implicitly using `maxRetries` = 0 correct? Is this intentional for the relayer? From my perspective we may want at least a couple retries in the scenario of reorg

x/contracts/ethwrapper/client.go
```diff
@@ -0,0 +1,236 @@
...
+
+// BlockNumber returns the latest block number.
+func (c *Client) BlockNumber(ctx context.Context) (uint64, error) {
+	var errs error
+	for i := range c.maxRetries {
+		for _, client := range c.clients {
+			switch bn, err := client.cli.BlockNumber(ctx); {
+			case err == nil:
+				c.logger.Debug("get block number succeeded", "attempt", i, "block", bn)
+				return bn - uint64(c.blockNumberDrift), nil
```
shaspitz, line 118: Simply subtracting 64 blocks from the latest block number will work (since two epochs is max finalization time), but sometimes the block will actually be finalized sooner. This could be observed using the "finalized" option https://ethereum.org/en/developers/docs/apis/json-rpc/, but this may be too big of a refactor, as I believe the geth client doesn't support this finalized option by default

aloknerurkar: I have managed to add this to our ethclient wrapper. Can you check again?... Also just doing it for L1

bridge/standard/pkg/node/node.go
```diff
@@ -399,3 +392,29 @@ func setupMetricsNamespace(namespace string) {
...
+// FinalizedBlockNumGetter gets the finalized block number from the Ethereum client.
+func finalizedBlockNumGetter(ctx context.Context, cli ethwrapper.EthClient) (uint64, error) {
+	client, ok := cli.(*ethclient.Client)
+	if !ok {
+		return cli.BlockNumber(ctx)
+	}
+	var blockResponse BlockResponse
+
+	// Call the eth_getBlockByNumber RPC with "finalized"
+	err := client.Client().CallContext(ctx, &blockResponse, "eth_getBlockByNumber", "finalized", false)
```
shaspitz, line 410: Looks good thanks!

shaspitz (approved):

PR #461, opened 2024-10-31
fix: change staking API to allow register as well as deposit

aloknerurkar:
## Describe your changes
Currently the registerAndStake API only allows providers to register a new stake. If the providers get slashed and their stake goes below minimum, there is no API to deposit additional stake. There doesnt seem to be any benefit of adding a different API for this as users don't care much about this. Evidence from the customers we talked to. So this PR changes the current API to register if this is a new stake, or deposit funds to the same stake if its already there.

Fixes # (issue)


p2p/pkg/rpc/provider/service.go
```diff
@@ -224,15 +226,23 @@ func (s *Service) RegisterStake(
...
-	}
+	var (
+		tx             *types.Transaction
+		blsPubkeyBytes []byte
+	)
 
-	tx, err := s.registryContract.RegisterAndStake(opts, blsPubkeyBytes)
+	switch _, err = s.registryContract.GetProviderStake(&bind.CallOpts{Context: ctx, From: s.owner}, s.owner); {
+	case err != nil:
+		blsPubkeyBytes, err = hex.DecodeString(strings.TrimPrefix(stake.BlsPublicKey, "0x"))
```
mrekucci, line 240: The original error is shadowed here.

p2p/pkg/rpc/provider/service_test.go
```diff
@@ -36,8 +37,16 @@ func (t *testRegistryContract) RegisterAndStake(opts *bind.TransactOpts, blsPubl
...
 }
 
+func (t *testRegistryContract) Stake(opts *bind.TransactOpts) (*types.Transaction, error) {
+	t.topup = opts.Value
+	return types.NewTransaction(1, common.Address{}, nil, 0, nil, nil), nil
+}
+
 func (t *testRegistryContract) GetProviderStake(_ *bind.CallOpts, address common.Address) (*big.Int, error) {
-	return t.stake, nil
+	if t.stake.Cmp(big.NewInt(0)) == 0 {
```
mrekucci: You can use `Sign()` method instead of `Cmp()` (it's more effective).

p2p/pkg/rpc/provider/service_test.go
```diff
@@ -36,8 +37,16 @@ func (t *testRegistryContract) RegisterAndStake(opts *bind.TransactOpts, blsPubl
...
+		return big.NewInt(0), errors.New("no stake")
```
mrekucci: Zero value of pointer is `nil` not `big.NewInt(0)`.

mrekucci (approved):
Apart from a few comments, LGTM!

p2p/pkg/rpc/provider/service_test.go
aloknerurkar: I am initializing it to big.NewInt(0). Which is why this check. I had to change this anyway as the contract returns 0 even if stake is not registered. So I have used a different contract function now.

PR #462, opened 2024-11-01
feat: updates autodeposit base amount to 300 gwei for window

ckartik:


aloknerurkar (commented):
I think its better to have this as an option while starting the script if people want to configure this. I am guessing they would want to.

shaspitz (approved):

aloknerurkar (approved):

Issue #500, opened 2024-11-25
Get rid of redundant mapping bidder registry


Mikelle:
Get rid of redundant bidder registration mapping in BidderRegistry.sol

PR #501, opened 2024-11-25
fix: deleted bidder registered mapping

Mikelle:
## Describe your changes

Got rid of redundant mapping in bidder registry contract

Fixes https://github.com/primev/mev-commit/issues/500


shaspitz:
Naïve questions - How would a provider check if a bidder is registered on-chain following this PR? Does a provider need to do this?

Mikelle:
> How would a provider check if a bidder is registered on-chain following this PR?

Provider just check for the bidder balance in the window. Current check is not very helpful, as bidder could withdraw all the funds and still be "registered"

> Does a provider need to do this?

No need to do this, as he checks for window balance only.

shaspitz (approved):

PR #523, opened 2024-12-09
feat: adds BLS signature verification for provider registration

ckartik:
This PR introduces updates to the smart contracts, testing infrastructure, and other components impacting provider flow, registration, and staking processes:

1. **Provider Registration**:
   - Providers must now sign their EOA account address during registration and include the signature in the StakeRequest.
     - The authentication of the EOA as the owner of the private key associated with the registered BLS public key.
     - Cryptographic linkage of the registered BLS key (used in relays) to the EOA, identifying the provider/builder on MEV-commit.

2. **Protobuf Updates**:
   - Updates were made to the protobufs to accommodate changes in the StakeRequest structure.

3. **P2P Nodes**:
   - Minor updates were made to aspects of P2P nodes to align with these changes.

These updates collectively strengthen the staking and registration process while improving integration with the relay and P2P components.

Fixes # (issue)

- [x] I have added tests that prove my fix is effective or that my feature works.
- [x] I have made corresponding changes to the documentation.


mrekucci (approved):

contracts/contracts/core/ProviderRegistry.sol
```diff
@@ -228,6 +234,35 @@ contract ProviderRegistry is
...
+        bytes calldata signature
+    ) external {
+        address provider = msg.sender;
+
+        require(providerRegistered[provider], "Provider not registered");
+        require(blsPublicKey.length == 48, "Public key must be 48 bytes");
+        require(signature.length == 96, "Signature must be 96 bytes");
+
+
+        bytes32 message = keccak256(abi.encodePacked(provider));
```
chrmatt, line 253: For such authentication schemes, it is always a good idea to include some context into the signed message to prevent replay attacks and to prevent misuse of the signature in another context.

Here, only the provider address is signed. If some other place also simply expects a signature on this address, an attacker can use the signature obtained here to cause harm in the other place. Better would be to bind the signed message to this context, e.g., by signing Hash("BLS Key addition to mev-commit provider registry" + "mainnet" / "testnet" + mev-commit version + provider address)

Once the context is established, one still should bind the signature to the particular instance of the protocol execution to prevent replay attacks. E.g., if a provider wants to move this BLS key to another provider account since somebody got access to the old one, the holder of the original account can register again with the same signature used previously. This can be prevented by some challenge-response scheme, or a time-stamp, e.g., in the form of the mev-commit chain height.

chrmatt (dismissed):
Verifying the BLS key is a good step! I suggest to not just sign the address though, as discussed in more detail below.

contracts/contracts/core/ProviderRegistry.sol
ckartik: That's a great suggestion! Thanks~

.github/workflows/ci.yml
```diff
@@ -23,7 +23,7 @@ concurrency:
 jobs:
   go-modules:
     name: Test and Build Go Modules
-    runs-on: ubuntu-24.04
+    runs-on: primev-ci-runner
```
mrekucci, line 26: Why is this changed? The same is below.

mrekucci (changes requested):

contracts/contracts/core/ProviderRegistry.sol
ckartik: I think for now this should be fine, we can update this in a later iteration. The provider address generated for the user in mev-commit is singularly used as an EOA account in mev-commit. We're also simply authenticating the BLS public key to be tied to the EOA account.  So I don't see this being an issue, at least for now. If we increase the use-cases for BLS signatures, we can move onto this approach.

mrekucci (approved):

PR #527, opened 2024-12-09
feat: use multiple RPCs for relayer

aloknerurkar:
## Describe your changes
Use multiple RPC URLs for bridge for failover incase one of them is down.

Fixes # (issue)

### DIFF
--- a/bridge/standard/pkg/node/node.go
+++ b/bridge/standard/pkg/node/node.go
@@ -37,7 +37,7 @@ type Options struct {
 	Logger                 *slog.Logger
 	HTTPPort               int
 	Signer                 keysigner.KeySigner
-	L1RPCURL               string
+	L1RPCURLs              []string
 	L1GatewayContractAddr  common.Address
 	SettlementRPCURL       string
 	SettlementContractAddr common.Address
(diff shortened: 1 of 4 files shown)


bridge/standard/pkg/node/node.go
```diff
@@ -100,7 +100,7 @@ func NewNode(opts *Options) (*Node, error) {
 		ctx, cancel := context.WithTimeout(context.Background(), 15*time.Second)
 		defer cancel()
 
-		client, err := ethclient.DialContext(ctx, opts.L1RPCURL)
+		client, err := ethclient.DialContext(ctx, opts.L1RPCURLs[0])
```
Mikelle, line 103: What is the point to submit an array, if we just take the first value?

Mikelle: Oh, we assume, that first one will work and in wrapped client use the array

Mikelle (approved):

bridge/standard/pkg/node/node.go
aloknerurkar: Yes, technically all work. If the first one doesnt work, we will fail in the initialization itself. This is mostly needed if we get rate-limited or if the RPC is down at some later point.

For the initial call we only do it to get the chain-id. We dont configure it in the CLI.

shaspitz (approved):

mrekucci (approved):

PR #541, opened 2024-12-20
feat: adds engineering script to monitor/track registries

ckartik:
Details in video here: https://www.loom.com/share/65dfb6b9ba23461d880561ca1d2bc812?sid=be0751dd-ff89-49c4-8c87-0d861e2146d7


aloknerurkar:
There is already a [dashboard](https://github.com/primev/mev-commit/tree/main/tools/dashboard).

Ideally you could have just added an extra channel to the existing event handlers and added some UI to the same place instead of having a separate tool.

Also the CI is failing because you need to use replace in go.mod to refer to the contracts-abi local module.

PR #543, opened 2024-12-23
fix: validator check lag

aloknerurkar:
## Describe your changes
Use the same lag as oracle while querying for opted-in validators

Fixes # (issue)

### DIFF
--- a/p2p/cmd/main.go
+++ b/p2p/cmd/main.go
@@ -431,10 +431,10 @@ var (
 		Category: categoryProvider,
 	})
 
-	optionLaggerdMode = altsrc.NewIntFlag(&cli.IntFlag{
-		Name:     "laggerd-mode",
+	optionLaggardMode = altsrc.NewIntFlag(&cli.IntFlag{
+		Name:     "laggard-mode",
 		Usage:    "No of blocks to lag behind for L1 chain when fetching validator duties",
-		EnvVars:  []string{"MEV_COMMIT_LAGGERD_MODE"},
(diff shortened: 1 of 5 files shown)


p2p/cmd/main.go
```diff
@@ -430,6 +430,14 @@ var (
 		Value:    30 * time.Second,
 		Category: categoryProvider,
 	})
+
+	optionLaggerdMode = altsrc.NewIntFlag(&cli.IntFlag{
+		Name:     "laggerd-mode",
```
Mikelle, line 435:
```suggestion
		Name:     "laggard-mode",
```

p2p/pkg/node/node.go
```diff
@@ -416,17 +417,24 @@ func NewNode(opts *Options) (*Node, error) {
...
-			CallOpts: bind.CallOpts{
-				From: opts.KeySigner.GetAddress(),
-			},
+		callOptsGetter := func() (*bind.CallOpts, error) {
+			blkNum, err := l1ContractRPC.BlockNumber(context.Background())
+			if err != nil {
+				return nil, err
+			}
+			currentBlkNum := big.NewInt(0).SetUint64(blkNum)
+			queryBlkNum := big.NewInt(0).Sub(currentBlkNum, big.NewInt(int64(opts.LaggeredMode)))
```
Mikelle, line 426: Could we initialize opts.laggersMode as big.Int, so we could avoid casting on every api call?

Mikelle (approved):

PR #558, opened 2025-01-08
fix: use residual percentage instead of decay

aloknerurkar:
## Describe your changes
Oracle has to send residual percentage for settlement.

## Issue ticket number and link

Fixes # (issue)

## Checklist before requesting a review

- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have made corresponding changes to the documentation

PR #559, opened 2025-01-08
feat: changes decayPercentage to residualAfterDecayPercentage

ckartik:
* Resolves inverted calculation of decay in the Oracle.
* Ensures value sent is the residual bid **after** decay, rather than decay.


mrekucci (commented):
Isn't this a duplicate of #558?

ckartik:
Yeah, I didn't see that one. Will close this in that case.

PR #560, opened 2025-01-08
feat: comprehensive overhaul of events for traceability of funds

ckartik:
* There's a lack of a comprehensive correlation-id off which, funds can be tracked in their movement across are protocol for a given Commitment.
* This PR introduces the use of commitment digest, as such a correlation ID to associate the waterfall like spitting of funds across multiple actors in the progression of the preconfirmation commitment lifecycle.


contracts/contracts/core/BidderRegistry.sol
```diff
@@ -179,6 +179,7 @@ contract BidderRegistry is
         uint256 amtMinusFeeAndDecay = decayedAmt - feeAmt;
 
         protocolFeeTracker.accumulatedAmount += feeAmt;
+        emit ProtocolFeeTransferred(commitmentDigest, feeAmt, protocolFeeTracker.recipient);
```
shaspitz, line 182: Fees are not necessarily transferred here, they're accumulated

```suggestion
        emit ProtocolFeeAccumulated(commitmentDigest, feeAmt, protocolFeeTracker.recipient);
```

contracts/contracts/core/ProviderRegistry.sol
```diff
@@ -117,16 +118,19 @@ contract ProviderRegistry is
         providerStakes[provider] -= residualAmt + penaltyFee;
 
         penaltyFeeTracker.accumulatedAmount += penaltyFee;
+        emit PenaltyFeeAllocated(commitmentDigest, penaltyFee, provider, penaltyFeeTracker.recipient);
```
shaspitz, line 121: nit
```suggestion
        emit PenaltyFeeAccumulated(commitmentDigest, penaltyFee, provider, penaltyFeeTracker.recipient);
```

Also might want to include the accumulated amount

contracts/contracts/core/ProviderRegistry.sol
```diff
@@ -98,6 +98,7 @@ contract ProviderRegistry is
      * @param residualBidPercentAfterDecay The residual bid percent after decay.
      */
     function slash(
+        bytes32 commitmentDigest,
```
shaspitz, line 101: Isn't this already emitted in the `SlashInitiated` event, during the calling function

contracts/contracts/core/BidderRegistry.sol
shaspitz: also should log accumulatedAmount

Evan-Kim2028:
how do we know all of these events will work as intended? Is there a way to test that the event will be triggered successfully with the intended output before committing all of these changes?

contracts/contracts/core/Oracle.sol
```diff
@@ -152,27 +152,27 @@ contract Oracle is OracleStorage, IOracle, Ownable2StepUpgradeable, UUPSUpgradea
...
     /**
      * @dev Internal function to process a commitment, either slashing or rewarding based on the commitment's state.
-     * @param commitmentIndex The id of the commitment to be processed.
+     * @param commitmentDigest The hash of the commitment to be processed.
      * @param isSlash Determines if the commitment should be slashed or rewarded.
      * @param residualBidPercentAfterDecay The residual bid percent after decay.
      */
     function _processCommitment(
-        bytes32 commitmentIndex,
+        bytes32 commitmentDigest,
```
aloknerurkar, line 160: Why is this changed? The caller of this function still uses commitmentIndex. Does this have to be renamed? We use the commitmentIndex when this is called from Oracle.

contracts/contracts/core/PreconfManager.sol
```diff
@@ -400,19 +400,19 @@ contract PreconfManager is
...
-     * @param commitmentIndex The hash of the commitment to be slashed.
+     * @param commitmentDigest The hash of the commitment to be slashed.
      * @param residualBidPercentAfterDecay The residual bid percent after decay.
      */
     function initiateSlash(
-        bytes32 commitmentIndex,
+        bytes32 commitmentDigest,
         uint256 residualBidPercentAfterDecay
     ) public onlyOracleContract whenNotPaused {
         OpenedCommitment storage commitment = openedCommitments[
```
aloknerurkar, line 410: This mapping also uses commitmentIndex

aloknerurkar (commented):
We should also evaluate how this affects gas costs for the transactions. Need to check gas used from the actual transactions before and after the change.

shaspitz:
> We should also evaluate how this affects gas costs for the transactions. Need to check gas used from the actual transactions before and after the change.

Gas cost for events should be pretty minimal compared to storage https://www.rareskills.io/post/ethereum-events#:~:text=Gas%20cost%20to%20emit%20Solidity%20event&text=The%20formula%20for%20how%20much,paid%20for%20each%20indexed%20parameter.

shaspitz:
> how do we know all of these events will work as intended? Is there a way to test that the event will be triggered successfully with the intended output before committing all of these changes?

To test the emittance of events, we use https://book.getfoundry.sh/cheatcodes/expect-emit in our foundry tests. You're correct in that PR doesn't add tests to confirm event emittance

contracts/contracts/core/Oracle.sol
ckartik: Changing this primarily to keep a consistent naming convention

contracts/contracts/core/PreconfManager.sol
ckartik: We should change the name in the mapping as well, they're the same value, I just think we should keep a consistent variable name

contracts/contracts/core/Oracle.sol
aloknerurkar: https://github.com/primev/mev-commit/blob/main/oracle/pkg/updater/updater.go#L407

What I meant is this argument seems to be commitmentIndex. I dont think this is commitmentDigest.

ckartik: Yeah, you're right. It's a hash of the digest+sig. https://github.com/primev/mev-commit/blob/50f6efefd0836cf7e790d59ded277a37e338196d/contracts/contracts/core/PreconfManager.sol#L624

PR #570, opened 2025-01-17
fix: autodeposit during node init

aloknerurkar:
## Describe your changes
The txmonitor should be started before we start making transactions. This component rate-limits txns based on account nonces. So if it is not started a higher nonce could be considered invalid.

### DIFF
--- a/p2p/pkg/node/node.go
+++ b/p2p/pkg/node/node.go
@@ -588,6 +588,7 @@ func NewNode(opts *Options) (*Node, error) {
 	}
 
 	ctx, cancel := context.WithCancel(context.Background())
+	nd.cancelFunc = cancel
 	healthChecker := health.New()
 
 	for _, s := range startables {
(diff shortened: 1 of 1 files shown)


p2p/pkg/node/node.go
```diff
@@ -603,6 +596,14 @@ func NewNode(opts *Options) (*Node, error) {
...
 
+	if opts.PeerType == p2p.PeerTypeBidder.String() && opts.AutodepositAmount != nil {
+		err = nd.autoDeposit.Start(ctx, nil, opts.AutodepositAmount)
+		if err != nil {
+			opts.Logger.Error("failed to start auto deposit tracker", "error", err)
+			return nil, errors.Join(err, nd.Close())
+		}
+	}
+
 	nd.cancelFunc = cancel
```
mrekucci, line 607: The cancel function should be used in all paths.

Mikelle (approved):

mrekucci (approved):

Issue #572, opened 2025-01-17
WIthdraw from Window lacks observability in edge case


ckartik:
```
function withdrawFromWindows(
        uint256[] calldata windows
    ) external nonReentrant whenNotPaused {
        uint256 currentWindow = blockTrackerContract.getCurrentWindow();
        uint256 totalAmount;

        uint256 len = windows.length;
        for (uint256 i = 0; i < len; ++i) {
            uint256 window = windows[i];
            require(window < currentWindow, WithdrawAfterWindowSettled(window, currentWindow));

            uint256 amount = lockedFunds[msg.sender][window];

            lockedFunds[msg.sender][window] = 0;
            maxBidPerBlock[msg.sender][window] = 0;

            (uint256 startBlock, uint256 endBlock) = WindowFromBlockNumber.getBlockNumbersFromWindow(window);

            for (uint256 blockNumber = startBlock; blockNumber <= endBlock; ++blockNumber) {
                usedFunds[msg.sender][uint64(blockNumber)] = 0;
            }

            emit BidderWithdrawal(msg.sender, window, amount);

            totalAmount += amount;
        }

        (bool success, ) = msg.sender.call{value: totalAmount}("");
        require(success, BidderWithdrawalTransferFailed(msg.sender, totalAmount));
    }
```

Observe above tha tin the case where the bidder is unable to successfully execute `(bool success, ) = msg.sender.call{value: totalAmount}("");` there's no event that corresponds to this, instead from the perspective of events, the bidders withdrawal was successful.

I useful mitigation to this is to move the bidderWithdrawl event to be fired after successful withdrawal of funds.

PR #573, opened 2025-01-17
chore: adds testing and ensures event emit matches state

ckartik:
Resolves issue #572


ckartik:
Closing, since the state and emitted events already reverts when the transaction fails.

PR #574, opened 2025-01-17
feat: adds observability into window deposits

ckartik:
* Adds viability into granular movement of funds into window
* Adds event that updates user about max-bid per block in a given window

### DIFF
--- a/contracts/contracts/core/BidderRegistry.sol
+++ b/contracts/contracts/core/BidderRegistry.sol
@@ -87,8 +87,7 @@ contract BidderRegistry is
         maxBidPerBlock[msg.sender][window] = newLockedFunds / WindowFromBlockNumber.BLOCKS_PER_WINDOW;
         
         emit BidderRegistered(msg.sender, newLockedFunds, window);
-        emit BidderDepositedForWindow(msg.sender, window, msg.value, newLockedFunds);
-        emit MaxBidForWindowUpdated(msg.sender, window, maxBidPerBlock[msg.sender][window]);
+        emit BidderDepositedForWindow(msg.sender, window, msg.value, newLockedFunds, maxBidPerBlock[msg.sender][window]);
     }
 
     /**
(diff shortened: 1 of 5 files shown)


contracts/contracts/core/BidderRegistry.sol
```diff
@@ -114,8 +116,11 @@ contract BidderRegistry is
             maxBidPerBlock[msg.sender][window] =
                 newLockedFunds /
                 WindowFromBlockNumber.BLOCKS_PER_WINDOW;
-
+            
             emit BidderRegistered(msg.sender, newLockedFunds, window);
+            emit BidderDepositedForWindow(msg.sender, window, amountToDeposit, newLockedFunds);
```
shaspitz, line 121: Why do we need to emit `BidderRegistered` right before this new event, `BidderDepositedForWindow`? They seem to emit the same data

ckartik: The new event includes amountToDesposit (e.g the amount being deposited by the user) The old one only included the aggregate funds in the window. Both are being emited for backwards compatibility.

shaspitz (approved):
Approved given we add a todo, or track the task of removing the old events once everyone's migrated to a new version

contracts/contracts/core/BidderRegistry.sol
```diff
@@ -114,8 +116,11 @@ contract BidderRegistry is
...
+            emit MaxBidForWindowUpdated(msg.sender, window, maxBidPerBlock[msg.sender][window]);
```
aloknerurkar, line 122: I am not sure why this separate event is needed. The max bid per block is function of window length and amount. So it makes sense to include it in the same event above.

contracts/contracts/core/BidderRegistry.sol
aloknerurkar: This event is used by the DepositManager to keep track of balances of the bidder. So if we want to use this new event, there will be changes required on mev-commit side as well.

ckartik:
We can make the new changes to mev-commit in follow on PRs, I'd rather keep contract code changes in a single PR. @aloknerurkar

contracts/contracts/core/BidderRegistry.sol
ckartik: Yeah since it's specific to the bidder makes sense to just consolidate

Issue #548, opened 2024-12-30
Create global application event bus


aloknerurkar:
We should create a global event bus package for mev-commit which can allow different packages to publish and subscribe to events. For eg:
- Peer disconnections
- Provider slashes for provider node as well as other bidder nodes
- Rewards
- Failed storing/opening commitments

Different sub-systems need to react to different events. So having a global event bus is a cleaner design than to do this specific to packages. This event bus can also be used to provide a new API for users to see these events in their infrastructure also.

For eg, provider slashing beyond certain point should lead to bidders re-configuring the encryption key and disconnecting from the provider. Also, currently if we fail to store commitment, there is no feedback for the provider nodes. This can happen due to balance being low or issues with the RPC endpoint. Both need to be addressed. The new API can allow users to see these events as well as other packages to react.

PR #575, opened 2025-01-21
feat: new notifications service

aloknerurkar:
## Describe your changes
New notifications service and new API for the same. Added some basic events and modified tests to ensure correctness.

#548 


p2p/pkg/notifications/notifications.go
```diff
@@ -0,0 +1,66 @@
...
+// Notifier is an interface that is used to notify about a notification. It will be
+// used by the publisher to notify the subscribers about the notification.
+type Notifier interface {
+	Notify(*Notification)
+}
+
+// Notifiee is an interface that is used to subscribe to a notification. It will be
+// used by the subscriber to subscribe to the notification.
+type Notifiee interface {
+	Subscribe(topics ...string) chan *Notification
```
mrekucci, line 30: The topics param should be typed (not string) so only topics enumerated above can be passed.

p2p/pkg/notifications/notifications.go
```diff
@@ -0,0 +1,66 @@
...
+	TopicPeerConnected    = "peer_connected"
+	TopicPeerDisconnected = "peer_disconnected"
+)
+
+// Notification is a struct that represents a notification. It has a Topic field
+// that represents the topic of the notification and a Value field that represents
+// the value of the notification. The Value field is a map[string]any, which means
+// that it can be any type.
+type Notification struct {
+	Topic string
```
mrekucci, line 17: This fields shouldn't be exported so they cannot be modified outside of the package.

p2p/pkg/notifications/notifications.go
```diff
@@ -0,0 +1,92 @@
+package notifications
```
mrekucci, line 1: I'd suggest to rename the package to `notification`.

mrekucci (approved):

Mikelle (approved):

p2p/pkg/notifications/notifications.go
aloknerurkar: Any specific reason? Usually I name package name as collection rather than single.

PR #576, opened 2025-01-22
feat: adds better tracking of funds going to treasury

ckartik:
* Improves traceability of funds going to treasury
* Improves slashing traceability based on commitment digest.


.gitmodules
```diff
@@ -19,9 +19,11 @@
 [submodule "contracts/lib/core"]
 	path = contracts/lib/core
 	url = https://github.com/symbioticfi/core
+	tag = 1.0.0-devnet.10
```
shaspitz, line 22: I don't think a .gitmodules file supports `tag` like this. See https://git-scm.com/docs/gitmodules. What are you attempting to do here?

contracts/contracts/core/BidderRegistry.sol
```diff
@@ -179,6 +179,7 @@ contract BidderRegistry is
         uint256 amtMinusFeeAndDecay = decayedAmt - feeAmt;
 
         protocolFeeTracker.accumulatedAmount += feeAmt;
+        emit FeePayout.FundsAccumulatedForTreasury(commitmentDigest, feeAmt, protocolFeeTracker.recipient, bidState.bidder, provider);
```
shaspitz, line 182: If adding this we should ensure  `protocolFeeTracker.accumulatedAmount` is emitted

contracts/contracts/core/ProviderRegistry.sol
```diff
@@ -117,6 +118,7 @@ contract ProviderRegistry is
         providerStakes[provider] -= residualAmt + penaltyFee;
 
         penaltyFeeTracker.accumulatedAmount += penaltyFee;
+        emit FeePayout.FundsAccumulatedForTreasury(commitmentDigest, penaltyFee, penaltyFeeTracker.recipient, bidder, provider);
```
shaspitz, line 121: Same comment as above w.r.t. `accumulatedAmount`

contracts/contracts/core/ProviderRegistry.sol
```diff
@@ -127,6 +129,7 @@ contract ProviderRegistry is
         }
 
         emit FundsSlashed(provider, residualAmt + penaltyFee);
+        emit FundsSlashedV2(provider, residualAmt + penaltyFee, commitmentDigest);
```
shaspitz, line 132: Why not just add a field to the existing `FundsSlashed` event?

.gitmodules
ckartik: Yeah looks like it.

contracts/contracts/core/ProviderRegistry.sol
ckartik: To maintain backwards compatibility

shaspitz (approved):
Approved given we add a todo, or track the task of removing the old events once everyone's migrated to a new version

contracts/contracts/core/ProviderRegistry.sol
aloknerurkar: There is a reference to this event in the dashboard. We will need to update that as well.

https://github.com/primev/mev-commit/blob/main/tools/dashboard/stathandler.go#L270

contracts/contracts/core/ProviderRegistry.sol
```diff
@@ -126,7 +128,7 @@ contract ProviderRegistry is
             bidderSlashedAmount[bidder] += residualAmt;
         }
 
-        emit FundsSlashed(provider, residualAmt + penaltyFee);
+        emit FundsSlashed(provider, residualAmt + penaltyFee, commitmentDigest);
```
aloknerurkar, line 131: So I thought what we discussed is emitting the new event and not changing the old one. If we change the old one I am not sure how backward compatibility works. For eg, if they listen to old contract events with the new ABI what would happen? Would it be able to fill only the relevant items?

contracts/contracts/interfaces/IProviderRegistry.sol
```diff
@@ -10,7 +10,7 @@ interface IProviderRegistry {
     event FundsDeposited(address indexed provider, uint256 amount);
 
     /// @dev Event emitted when funds are slashed
-    event FundsSlashed(address indexed provider, uint256 amount);
+    event FundsSlashed(address indexed provider, uint256 amount, bytes32 indexed commitmentDigest);
```
aloknerurkar, line 13: If we add a new field which is indexed I am not sure if it will be able to Watch/Filter this event if we listen for it for block numbers before the contract update was made. This might be a problem for indexing the chain data using other tools.

PR #584, opened 2025-01-26
chore: fix ci pipeline

ckartik:
* Updates to [this](https://github.com/OpenZeppelin/openzeppelin-upgrades/releases/tag/%40openzeppelin%2Fupgrades-core%401.42.1) new release that fixes the build issues
* Adds pausable_init call for vanilla registry (seemed to be missing) @shaspitz - can you take a look at this?

### DIFF
--- a/contracts/contracts/validator-registry/VanillaRegistry.sol
+++ b/contracts/contracts/validator-registry/VanillaRegistry.sol
@@ -46,6 +46,7 @@ contract VanillaRegistry is IVanillaRegistry, VanillaRegistryStorage,
         uint256 _slashingPayoutPeriodBlocks,
         address _owner
     ) external initializer {
+        __Pausable_init();
         _setMinStake(_minStake);
         _setSlashOracle(_slashOracle);
         _setUnstakePeriodBlocks(_unstakePeriodBlocks);


contracts/contracts/validator-registry/VanillaRegistry.sol
```diff
@@ -46,6 +46,7 @@ contract VanillaRegistry is IVanillaRegistry, VanillaRegistryStorage,
         uint256 _slashingPayoutPeriodBlocks,
         address _owner
     ) external initializer {
+        __Pausable_init();
```
aloknerurkar, line 49: I would suggest moving this to a separate change and unblock the CI first. I think this change will need vanilla registry contract upgrade which is on L1. I am not sure how to handle this.

mrekucci (approved):
Except the change in the vanilla registry, LGTM!

contracts/contracts/validator-registry/VanillaRegistry.sol
ckartik: This was also causing the CI to fail, need to add it in or the tests compain.

ckartik: I checked with @shaspitz and he Ok'd the change [here](https://primevworkspace.slack.com/archives/C067BA9C7SS/p1737950804989999?thread_ts=1737948734.259909&cid=C067BA9C7SS)

PR #586, opened 2025-01-28
feat: zk shared key

Mikelle:

Fixes # (issue)


p2p/pkg/crypto/ecdh.go
```diff
@@ -0,0 +1,111 @@
...
+	sk = &fr.Element{}
+	_, err = sk.SetRandom()
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to generate random secret: %w", err)
+	}
+
+	// 2) Retrieve the G1 generator (1,2) from the bn254 package
+	var g1Aff bn254.G1Affine
+	g1Aff.X.SetOne()
+	g1Aff.Y.SetUint64(2)
```
chrmatt, line 25: Why is this hard-coded instead of using library function? See https://github.com/consensys/gnark-crypto/blob/v0.16.0/ecc/bn254/bn254.go#L151

contracts/contracts/core/PreconfManager.sol
```diff
@@ -650,23 +636,93 @@ contract PreconfManager is
...
+            getPreConfHash(params.bidHash, params.bidSignature, params.zkProof);
+    }
+
+    /**
+     * @notice Verifies the sigma-proof that:
+     *  (1) providerPub = g^sk
+     *  (2) sharedSecret = bidPub^sk
+     * using the non-interactive challenge c = H(...).
+     *
+     * The proof is {c, z}, plus we have inputs {providerPub, bidPub, sharedSecret}.
```
chrmatt, line 679: Would be good to recall here what zkProof[0], zkProof[1], etc., are.

p2p/pkg/preconfirmation/encryptor/encryptor.go
```diff
@@ -87,11 +87,22 @@ func NewEncryptor(ks keysigner.KeySigner, store Store, chainID *big.Int, preconf
...
 	if bid.TxHash == "" || bid.BidAmount == "" || bid.BlockNumber == 0 {
 		return nil, nil, ErrMissingRequiredFields
 	}
-
+	var (
+		sk  *fr.Element
+		pk  *bn254.G1Affine
+		err error
+	)
+	if bid.NikePublicKey == nil {
```
aloknerurkar, line 99: Why is this check required? We will always generate a new one no?

aloknerurkar (approved):
Apart from the minor comments LGTM!

p2p/pkg/crypto/ecdh.go
Mikelle: It’s hardcoded to use the same generators as on the Solidity side. Otherwise, I couldn’t verify the proof.

chrmatt: OK, makes sense. But then I think the comment "Retrieve the G1 generator (1,2) from the bn254 package" is misleading.

p2p/pkg/crypto/ecdh.go
```diff
@@ -0,0 +1,111 @@
...
+
+	var pub bn254.G1Affine
+	consumed, err := pub.SetBytes(data)
+	if err != nil {
+		return nil, fmt.Errorf("SetBytes error: %w", err)
+	}
+
+	// 2) Ensure exactly 96 bytes were consumed
+	if consumed != bn254.SizeOfG1AffineUncompressed {
+		return nil, fmt.Errorf("unexpected consumed bytes. got=%d want=%d",
```
mrekucci, line 68: Avoid the period in the error message.

p2p/pkg/crypto/zkproof.go
```diff
@@ -0,0 +1,184 @@
...
+	providerPubX, providerPubY := AffineToBigIntXY(providerPub)
+	bidPubX, bidPubY := AffineToBigIntXY(bidPub)
+	sharedSecX, sharedSecY := AffineToBigIntXY(sharedSec)
+	aX, aY := AffineToBigIntXY(a)
+	aX2, aY2 := AffineToBigIntXY(a2)
+
+	// 1) Flatten in the same order as abi.encodePacked.
+	//    We'll manually append each big.Int as 32-byte big-endian,
+	//    then the final keccak256 is your computedChallenge.
+	var buf []byte
```
mrekucci, line 148: If the length is known, the `buf` should be allocated along with it.

p2p/pkg/p2p/libp2p/providerkeys.go
```diff
@@ -1,23 +1,27 @@
...
 
+	"github.com/consensys/gnark-crypto/ecc/bn254"
+	"github.com/consensys/gnark-crypto/ecc/bn254/fr"
 	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/ethereum/go-ethereum/crypto/ecies"
+	p2pcrypto "github.com/primev/mev-commit/p2p/pkg/crypto"
 	"github.com/primev/mev-commit/p2p/pkg/p2p"
 )
 
 type Store interface {
```
mrekucci, line 14: Avoid Get prefix: https://go.dev/doc/effective_go#Getters

mrekucci (approved):
Apart from the comments, LGTM!

contracts/contracts/core/PreconfManager.sol
```diff
@@ -200,7 +212,7 @@ contract PreconfManager is
      * @param decayStartTimeStamp The start time of the decay
      * @param decayEndTimeStamp The end time of the decay
      * @param bidSignature The signature of the bid
-     * @param sharedSecretKey The shared secret key
+     * @param zkProof The zk proof
```
ckartik, line 215:
```suggestion
     * @param zkProof The zk proof is an array of c and z, representing the challenge and z = k - c*sk
```

ckartik: technically I guess that isn't the case, since you're wrapping a bunch of data into that array.

chrmatt (approved):
LGTM

PR #590, opened 2025-02-06
feat: added slash amount

Mikelle:
## Describe your changes

This PR introduces a comprehensive update to how bid slashing is handled throughout the system. In particular, it adds an explicit `slashAmt` parameter to bids and related commitment processing, ensuring that the amount to be slashed from a provider’s stake is clearly defined and propagated through the entire preconfirmation flow.

---

### 1. PreconfManager Updates

- **EIP-712 Type Hash Revision:**  
  The bid type hash now includes the `slashAmt` field. For example, the type string was extended from:
  ```solidity
  PreConfBid(string txnHash,string revertingTxHashes,uint256 bidAmt,uint64 blockNumber,uint64 decayStartTimeStamp,uint64 decayEndTimeStamp,uint256 bidderPKx,uint256 bidderPKy)
  ```
  to
  ```solidity
  PreConfBid(string txnHash,string revertingTxHashes,uint256 bidAmt,uint64 blockNumber,uint64 decayStartTimeStamp,uint64 decayEndTimeStamp,uint256 slashAmt,uint256 bidderPKx,uint256 bidderPKy)
  ```

This ensures that the bid’s slashing amount is an integral part of the signed data.

#### Bid Hash and Commitment Digest
- All functions that compute bid hashes (e.g. `getBidHash`) and commitment digests have been updated to accept and incorporate `slashAmt` as a parameter. This change propagates the new field through bid verification and commitment creation.

#### Internal Logic Adjustments
- Functions such as `_getBidderAddressAndCommitmentDigest` was introduced to avoid `stack too deep` issue and now forward `slashAmt` to the bid verification logic.

#### Slash Function Overhaul
The `slash` function now accounts for the explicit `slashAmt` when calculating the amounts to deduct from a provider’s stake. The new logic:
- **Computes the residual amount** based on the decayed bid amount and fee percentage.
- **Incorporates `slashAmt`** when determining the bidder’s payout and the total amount to slash.
- **Emits events and performs additional checks** if the provider’s stake is insufficient.

#### P2P and API Modifications
- Protobuf definitions (in bidder and provider API specs) and the corresponding OpenAPI (Swagger) files have been updated to include the new `slash_amount` field.


aloknerurkar (commented):
We should add integration test for this. You can either modify the existing one or create a separate one for this.

aloknerurkar (approved):

PR #592, opened 2025-02-07
feat: adds a pointservice

ckartik:
Add Points Accrual Service with Event Subscription, API Endpoints, and Initial SQLite DB
output

**Summary**  
- Introduces the `PointsService` managing a periodic points accrual routine (`StartPointsRoutine`) and event subscription for validator opt-ins/opt-outs.  
- Provides an HTTP API (`PointsAPI`) for querying points totals and block info.  
- Implements a basic SQLite data model with `validator_records` and `last_processed_block` tables, plus unit tests for `computePointsForMonths`.  

**Next Steps**  
- **Migrate** the DB layer from SQLite to Postgres.  
- **Move** the service into our production infrastructure codebase, ensuring proper configuration (e.g., environment variables, container orchestration, etc.).


tools/points-service/main.go
```diff
@@ -0,0 +1,529 @@
...
+	if rowsErr := rows.Err(); rowsErr != nil {
+		err = fmt.Errorf("rows iteration error: %w", rowsErr)
+	}
+
+	return err
+}
+
+// StartPointsRoutine spawns a background goroutine that runs every 'interval'
+// and calls updatePoints with the latest block.
+func StartPointsRoutine(db *sql.DB, logger *slog.Logger, interval time.Duration, ethClient *ethclient.Client) {
```
aloknerurkar, line 178: This routine can be improved. First we should pass in a context to handle clean shutdowns. Second the initial run and ticker runs can be combined. Also in case of errors we should retry with smaller delay.

```suggestion
func StartPointsRoutine(ctx context.Context, db *sql.DB, logger *slog.Logger, interval time.Duration, ethClient *ethclient.Client) {

    ticker := time.NewTicker(interval)
    
    for {
        latestBlock, err := ethClient.BlockByNumber(ctx, nil)
	if err != nil {
		logger.Error("cannot fetch latest block", "error", err)
		ticker.Reset(errorDelay)
	} else {
		currBlockNum := latestBlock.NumberU64()
		// Ideally ctx should be passed in this function as well and checked in the update function.
		if err := updatePoints(db, logger, currBlockNum); err != nil {
			logger.Error("initial points accrual run failed", "error", err)
			ticker.Reset(errDelay)
		} else {
			logger.Info("initial points accrual run completed successfully")
			ticker.Reset(interval)
		}
	}
	
	select {
	    case <-ticker.C:
	    case <-ctx.Done():
	        return
	}
    }
```

tools/points-service/main.go
```diff
@@ -0,0 +1,529 @@
...
+
+type PointsService struct {
+	logger    *slog.Logger
+	db        *sql.DB
+	ethClient *ethclient.Client
+	block     uint64
+}
+
+func (ps *PointsService) LastBlock() (uint64, error) {
+	return ps.block, nil
```
aloknerurkar, line 230: This should be stored in the DB.

tools/points-service/main.go
```diff
@@ -0,0 +1,529 @@
...
+	rwLock.RLock()
+	defer rwLock.RUnlock()
+
+	var existingAdder string
+	err := db.QueryRow(`
+        SELECT adder FROM events
+        WHERE pubkey = ? AND opted_out_block IS NULL
+    `, pubkey).Scan(&existingAdder)
+
+	// if pubkey is active under a different adder, ignore
```
aloknerurkar, line 248: This can be avoided by just checking the error on Insert. I think sql should give you a specific error for unique key violation.

tools/points-service/main.go
```diff
@@ -0,0 +1,529 @@
...
+			// 4. Load ABIs
+			contractABIs, err := getContractABIs()
+			if err != nil {
+				return fmt.Errorf("failed to get contract ABIs: %w", err)
+			}
+			listener := events.NewListener(logger, contractABIs...)
+
+			// example addresses
+			testnetContracts := []common.Address{
+				common.HexToAddress("0xEDEDB8ed37A43Fd399108A44646B85b780D85DD4"), // AVS
```
aloknerurkar, line 329: This should use constants from the contracts-abi pkg. If the addresses change, this would be auto updated.

tools/points-service/main.go
```diff
@@ -0,0 +1,529 @@
...
+			}
+
+			// 6. Subscribe
+			sub, err := listener.Subscribe(handlers...)
+			if err != nil {
+				return fmt.Errorf("subscribe error: %w", err)
+			}
+			defer sub.Unsubscribe()
+
+			// 7. Start daily routine to recompute points
```
aloknerurkar, line 476: This service does not handle clean shutdown. Check other services to see how to do that. There should be some context and also signals (SIGINT/SIGTERM) should be handled. All the other services should have this functionality.

x/contracts/events/events.go
```diff
@@ -14,10 +14,15 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 )
 
+// HasRawLog is an optional interface you can implement on your event struct
+// so we can store log metadata (e.g., block number) after unpacking.
+type HasRawLog interface {
```
aloknerurkar, line 19: I dont see where this is used?

x/contracts/events/publisher/http.go
```diff
@@ -76,29 +75,43 @@ func (h *httpPublisher) Start(ctx context.Context, contracts ...common.Address)
...
+					for startBlock <= blockNumber {
+						endBlock := startBlock + 5000 - 1
+						if endBlock > blockNumber {
+							endBlock = blockNumber
+						}
+
+						q := ethereum.FilterQuery{
+							FromBlock: big.NewInt(int64(startBlock)),
+							ToBlock:   big.NewInt(int64(endBlock)),
+							Addresses: contracts,
```
aloknerurkar, line 90: Where are we adding to this dynamically?

ckartik: Oh sorry those changes were still local, pushing them now.

tools/points-service/main.go
ckartik: The reason we have this because teh adder pubkey is not unqiue, it's adder, piubkey and opted in block, therefore sql won't throw an error

tools/points-service/api.go
```diff
@@ -0,0 +1,316 @@
...
+	p.logger.Info("Starting External Points API", slog.String("addr", addr))
+	if err := srv.ListenAndServe(); err != nil && err != http.ErrServerClosed {
+		return err
+	}
+	return nil
+}
+
+// HealthCheck endpoint checks if the API is alive
+// GET /health
+func (p *PointsAPI) HealthCheck(w http.ResponseWriter, r *http.Request) {
```
aloknerurkar, line 60: If you want to add a health check, ideally you should ensure the points routine is running, subscriber for events is running etc.

contracts-abi/script.sh
```diff
@@ -46,9 +46,19 @@ extract_and_save_abi "$BASE_DIR/out/MevCommitMiddleware.sol/MevCommitMiddleware.
...
 
 ABI_DIR="./abi"
-GO_CODE_BASE_DIR="./clients"
+# Create external-abi directory
+EXTERNAL_ABI_DIR="./external-abi"
+mkdir -p "$EXTERNAL_ABI_DIR"
+
+# Download the Vault.json
+curl -o "$EXTERNAL_ABI_DIR/vault.json" \
+  "https://raw.githubusercontent.com/symbioticfi/core/main/out/Vault.sol/Vault.json"
```
aloknerurkar, line 58: We should point to some tagged version rather than main.

tools/points-service/main.go
```diff
@@ -0,0 +1,646 @@
...
+			// 4. Load ABIs
+			contractABIs, err := getContractABIs()
+			if err != nil {
+				return fmt.Errorf("failed to get contract ABIs: %w", err)
+			}
+			listener := events.NewListener(logger, contractABIs...)
+
+			// Example addresses
+			testnetContracts := []common.Address{
+				common.HexToAddress(config.HoleskyContracts.ValidatorOptInRouter),
```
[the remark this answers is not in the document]
ckartik: Yeah might be worth just passing it in as a env/CLI flag.

tools/points-service/api.go
```diff
@@ -0,0 +1,403 @@
...
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(status)
+	if err := json.NewEncoder(w).Encode(data); err != nil {
+		log.Printf("failed to encode JSON: %v", err)
+	}
+}
+
+// GetAnyPointsForAddress sums up points for the given address:
+//   - points_accumulated if `opted_out_block` is NULL
+//   - pre_six_month_points if `opted_out_block` is NOT NULL
```
aloknerurkar, line 360: Comment needs to be updated?

tools/points-service/main.go
```diff
@@ -0,0 +1,758 @@
...
+        id INTEGER PRIMARY KEY AUTOINCREMENT,
+        pubkey TEXT NOT NULL,
+        adder TEXT NOT NULL,
+        vault TEXT,
+        registry_type TEXT CHECK (registry_type IN ('vanilla', 'symbiotic', 'eigenlayer')),
+        event_type TEXT,
+        opted_in_block BIGINT NOT NULL,
+        opted_out_block BIGINT,
+        points_accumulated BIGINT DEFAULT 0,
+		pre_cliff_points BIGINT DEFAULT 0,
```
aloknerurkar, line 51: Fix indentation

tools/points-service/main.go
```diff
@@ -0,0 +1,758 @@
...
+)
+
+// ~216000 blocks is roughly one month on Ethereum (~2s block time)
+var blocksInOneMonth = int64(216000)
+
+var (
+	rwLock sync.RWMutex
+
+	createTableValidatorRecordsQuery = `
+    CREATE TABLE IF NOT EXISTS validator_records (
```
aloknerurkar, line 41: Use only one indentation for all. The create table queries use different indentation than the other queries.

tools/points-service/readme.md
```diff
@@ -0,0 +1,47 @@
+
+## Vanilla Registry Events
```
aloknerurkar, line 2: I think we have added more events no?

This is the pain with having READMEs. Either update it or remove it.

tools/points-service/api.go
```diff
@@ -0,0 +1,403 @@
...
+}
+
+func (p *PointsAPI) StartAPIServer(ctx context.Context, addr string) error {
+	r := mux.NewRouter()
+	r.HandleFunc("/health", p.HealthCheck).Methods("GET")
+	r.HandleFunc("/last_block", p.GetLastBlock).Methods("GET")
+	r.HandleFunc("/{receiver_type}/{receiver_address}", p.RecomputePointsForAddress).Methods("GET")
+	r.HandleFunc("/stats", p.GetTotalPointsStats).Methods("GET")
+	r.HandleFunc("/all", p.GetAllPoints).Methods("GET")
+
```
aloknerurkar, line 32: I think we also need an API to assign points manually no? This should be authenticated API. Look at oracle. There is already an example for this. Maybe this can be done in a separate PR also.

aloknerurkar (commented):
Apart from the nit picks LGTM!

I would still be more comfortable if someone else also takes a look, make sure I havent missed anything. Maybe @shaspitz

tools/points-service/api.go
```diff
@@ -0,0 +1,403 @@
...
+	}
+
+	if receiverType != "operator" {
+		resp := map[string]interface{}{
+			"address":      receiverAddr,
+			"receiver":     receiverType,
+			"block_number": blockNum,
+			"points":       float64(0),
+		}
+		writeJSON(w, resp, http.StatusOK)
```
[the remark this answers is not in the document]
ckartik: No it's a requirement from symbiotic API spect, they want results for non-operator types. See [here](https://symbioticfi.notion.site/External-Points-API-Specification-16981c079c178059b6b7e9740cf987a8).

tools/points-service/api.go
```diff
@@ -0,0 +1,403 @@
...
+		return
+	}
+
+	totalPoints, err := p.calculatePointsForSymbioticOperator(receiverAddr, blockNum)
+	if err != nil {
+		http.Error(w, "DB query failed", http.StatusInternalServerError)
+		return
+	}
+	if totalPoints == 0 {
```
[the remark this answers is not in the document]
ckartik: No, there could still be a zero, if the address is not in our DB. Hence no sign up bonus.

tools/points-service/api.go
ckartik: Yeah I don't think we should block on that for now. We'd also have access to the DB, so we can always add entries manually.

ckartik:
image

tools/points-service/readme.md
```diff
@@ -0,0 +1,68 @@
...
+
+```solidity
+event Staked(
+    address indexed sender,     // The account that called stake()
+    address withdrawalAddress,  // The address that can withdraw/unstake 
+    bytes pubkey,              // The validator's BLS public key
+    uint256 amount             // Amount of ETH staked
+);
+```
+
```
shaspitz, line 20: What about `Unstaked(msg.sender, validator.withdrawalAddress, pubKey, validator.balance)` and/or `StakeWithdrawn(msg.sender, expectedWithdrawalAddress, pubKey, balance)`?

tools/points-service/readme.md
```diff
@@ -0,0 +1,68 @@
...
+```solidity
+event ValRecordAdded(
+    bytes pubkey,              // The validator's BLS public key
+    address operator,          // The operator registering the validator
+    address vault,             // The vault securing the validator
+    uint256 position           // Position in the validator set
+);
+```
+
```
shaspitz, line 35: What about `ValidatorDeregistrationRequested` and/or `ValRecordDeleted`?

tools/points-service/main.go
```diff
@@ -0,0 +1,778 @@
...
+	if err == nil && existingAdder != "" && existingAdder != adder {
+		logger.Warn("pubkey already opted in by a different adder",
+			"pubkey", pubkey, "existing_adder", existingAdder, "new_adder", adder)
+		return
+	}
+
+	_, err = db.Exec(`
+        INSERT INTO validator_records (
+            pubkey, adder, vault, registry_type, event_type, 
+            opted_in_block, opted_out_block, points_accumulated
```
shaspitz, line 426: This data model assumes a single validator will have one of each `opted_in_block` and `opted_out_block`. If they re-opt in a second time after opting in/out once, would a new row be created? Ie. how does the data model handle for this edge case? 

Note our contracts allow these sort of registration "cycles" for all actors. Ex: https://github.com/primev/mev-commit/blob/9c22b771b2792356ddd169bba07dca4ae8d581a1/contracts/test/validator-registry/middleware/MevCommitMiddlewareTest.sol#L1035

tools/points-service/api.go
```diff
@@ -0,0 +1,403 @@
...
+	defer rows.Close()
+
+	for rows.Next() {
+		var (
+			vaultAddr     string
+			regType       string
+			optedInBlock  uint64
+			optedOutBlock sql.NullInt64
+		)
+		if scanErr := rows.Scan(&vaultAddr, &regType, &optedInBlock, &optedOutBlock); scanErr != nil {
```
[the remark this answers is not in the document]
ckartik: I don't understand, we're just scanning values here. If you mean can we have multiple entries with the same pubkey opted in multiple times, yes!

tools/points-service/api.go
```diff
@@ -0,0 +1,403 @@
...
+		} else {
+			totalPoints += recomputedPoints
+		}
+	}
+	if rowsErr := rows.Err(); rowsErr != nil {
+		return 0, rowsErr
+	}
+	return totalPoints, nil
+}
+func (p *PointsAPI) GetTotalPointsStats(w http.ResponseWriter, r *http.Request) {
```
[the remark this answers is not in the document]
ckartik: GetTotalPointsStats Is used to compute for a target blocknumber for symbiotic

tools/points-service/main.go
ckartik: Yes a new row is created.

tools/points-service/api.go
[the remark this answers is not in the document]
shaspitz: Was in reference to all the logic below this comment

shaspitz (approved):

PR #596, opened 2025-02-11
fix: use separate chain id for mock l1

shaspitz:
## Describe your changes

Currently the mock l1 used for integration tests has the same chain id as mev-commit testnet (17864). This PR changes the mock l1 to have a more realistic chain ID which is that of holesky (17000). 

### DIFF
--- a/infrastructure/nomad/playbooks/templates/jobs/mock-l1.nomad.j2
+++ b/infrastructure/nomad/playbooks/templates/jobs/mock-l1.nomad.j2
@@ -85,11 +85,11 @@ job "{{ job.name }}" {
         data = <<-EOH
           #!/usr/bin/env bash
 
-          # Mutate genesis file to only have a single alloc to contract deployer
+          # Mutate genesis file to only have a single alloc to contract deployer, and chainID of holesky (17000)
           ${GETH_BIN} --dev dumpgenesis > local/default_genesis.json
           jq -s '
             .[0].alloc = .[1].alloc
-            | .[0].config.chainId = .[1].config.chainId
+            | .[0].config.chainId = 17000
             | .[0]
           ' local/default_genesis.json local/genesis_{{ env }}-{{ version }}.json > local/genesis.json
           ${GETH_BIN} --datadir local/data init local/genesis.json


infrastructure/nomad/playbooks/templates/jobs/mock-l1.nomad.j2
```diff
@@ -85,11 +85,11 @@ job "{{ job.name }}" {
         data = <<-EOH
           #!/usr/bin/env bash
 
-          # Mutate genesis file to only have a single alloc to contract deployer
+          # Mutate genesis file to only have a single alloc to contract deployer, and chainID of holesky (17000)
           ${GETH_BIN} --dev dumpgenesis > local/default_genesis.json
           jq -s '
             .[0].alloc = .[1].alloc
-            | .[0].config.chainId = .[1].config.chainId
+            | .[0].config.chainId = 17000
```
aloknerurkar, line 92: @mrekucci Is this right? Or do we need to change this in some other files?

shaspitz: AFAIK an eth chain's chainID is set in the `genesis.json` upon calling `geth init`

aloknerurkar: Basically I just want to know if this is the best place to override this. So I am guessing this value right now is picked up from some config values we set. Maybe we need to update that value for this particular profile. So purely nomad implementation doubt.

mrekucci: This will pin the chain ID for other environments as well, we should create a new profile in the `environments.yml` file and use it for ci/integration tests.

shaspitz: Does the mock l1 correspond to an environment in the way you've designed it? Doesn't seem so. The curren change would only affect the mock l1

shaspitz: Following up here - 

> This will pin the chain ID for other environments as well

^ This is my intention in this PR. The mock-l1 should never have a chain id other than `17000`, `1`, etc. The current behavior where the mock l1 has a chain id of `17864`, is incorrect. And likely a symptom of copying config from our mev-commit chain setup somewhere. 

Adding an entry to the `environments.yml` doesn't seem appropriate here as the mock l1 is not it's own environment, It's a nomad job. 

If more desirable, I can set this chain id as an env var in the `profiles.yml` under the `mock_l1_job`. Would this suffice @mrekucci?

PR #597, opened 2025-02-11
feat: add stake manager

### LINKED ISSUE #549: Implement StakeManager
Similar to the current DepositManager, which is used by the provider to check balance of bidders, we also need a StakeManager which tracks the stake of providers.

The StakeManager needs to do the following:

1. Get minimum stake in the start and track if this value changes using events
2. Respond to GetStake queries to other packages without querying the RPC but by tracking ProviderStake and Slash events. This would prevent the additional RPC call and also allow us to generate events when provider stake becomes low.

aloknerurkar:
## Describe your changes
Addresses #549 

Fixes #549 

### DIFF
--- a/p2p/pkg/stakemanager/stakemanager.go
+++ b/p2p/pkg/stakemanager/stakemanager.go
@@ -264,6 +264,7 @@ func (sm *StakeManager) GetStake(ctx context.Context, provider common.Address) (
 	_, found := sm.unstakeReqs.Get(provider)
 	sm.stakeMu.RUnlock()
 	if found {
+		sm.logger.Debug("provider is in unstake requests, returning zero stake")
 		return big.NewInt(0), nil
 	}
 
--- a/p2p/pkg/stakemanager/stakemanager.go
+++ b/p2p/pkg/stakemanager/stakemanager.go
@@ -260,7 +260,7 @@ func (sm *StakeManager) Start(ctx context.Context) <-chan struct{} {
 	return doneChan
 }
 
-func (sm *StakeManager) GetStake(provider common.Address) (*big.Int, error) {
+func (sm *StakeManager) GetStake(ctx context.Context, provider common.Address) (*big.Int, error) {
 	sm.stakeMu.RLock()
 	_, found := sm.unstakeReqs.Get(provider)
 	sm.stakeMu.RUnlock()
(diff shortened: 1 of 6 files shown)


p2p/pkg/stakemanager/stakemanager.go
```diff
@@ -0,0 +1,308 @@
...
+	"github.com/ethereum/go-ethereum/common"
+	lru "github.com/hashicorp/golang-lru/v2"
+	providerregistry "github.com/primev/mev-commit/contracts-abi/clients/ProviderRegistry"
+	"github.com/primev/mev-commit/p2p/pkg/notifications"
+	"github.com/primev/mev-commit/x/contracts/events"
+	"golang.org/x/sync/errgroup"
+)
+
+type ProviderRegistryContract interface {
+	GetProviderStake(*bind.CallOpts, common.Address) (*big.Int, error)
```
mrekucci, line 21: Drop the `Get` prefix.

p2p/pkg/stakemanager/stakemanager.go
```diff
@@ -0,0 +1,308 @@
...
+	minStake         atomic.Pointer[big.Int]
+	logger           *slog.Logger
+}
+
+func NewStakeManager(
+	owner common.Address,
+	evtMgr events.EventManager,
+	providerRegistry ProviderRegistryContract,
+	notifier notifications.Notifier,
+	logger *slog.Logger,
```
mrekucci, line 42: nit: logger (also context) should be the first param.

p2p/pkg/stakemanager/stakemanager.go
```diff
@@ -0,0 +1,308 @@
...
+		select {
+		case <-egCtx.Done():
+			sm.logger.Info("event subscription context done")
+			return nil
+		case err := <-sub.Err():
+			return fmt.Errorf("error in event subscription: %w", err)
+		}
+	})
+
+	eg.Go(func() error {
```
Mikelle, line 110: Why do we handle all the events in a single goroutine? Probably FundsSlashed event will be more frequent, than other events

mrekucci (commented):
Aside from the comments, LGTM!

p2p/pkg/stakemanager/stakemanager.go
aloknerurkar: I did not want to add more routines. I had planned to use buffered channels. Fixed it.

Mikelle (approved):

mrekucci (approved):

PR #598, opened 2025-02-11
feat: added notifications for validators opted in

Mikelle:
## Describe your changes

This PR introduces scheduling notifications for validator opt-ins, and running periodic epoch-based processing.

### Topic Update  
In `p2p/pkg/notifications/notifications.go`, a small addition was made (e.g., to support additional notification topics such as `TopicValidatorOptedIn`).  

### Service Initialization  
In `p2p/pkg/rpc/validator/service.go`, the `NewService` function has been refactored to:  
- Return an error along with the service instance.  
- Accept an additional `notifier` parameter to manage notifications.  

### Genesis Time Fetching  
The service now fetches the genesis time from the Beacon API (`/eth/v1/beacon/genesis`). This timestamp is crucial for calculating slot start times.  

### Notification Scheduling  
A new method, `scheduleNotificationForSlot`, has been implemented. It:  
- Computes the start time of a slot based on the fetched genesis time and a defined slot duration.  
- Schedules a notification to be sent before the slot starts (using a configurable notify offset).  
- Sends a notification via the provided notifier, including key details such as epoch, slot, and BLS key.  

### Epoch Processing & Cron Job  
- The `processEpoch` method now processes proposer duties, checks each validator’s opt-in status, and schedules notifications accordingly.  
- A new `Start` method runs a background cron job that:  
  - Calculates the upcoming epoch based on the genesis time and epoch duration.  
  - Waits until the appropriate time (using a configurable fetch offset) to fetch and process the epoch.  
  - Logs the scheduling and processing of epochs.  


p2p/pkg/rpc/validator/service.go
```diff
@@ -216,3 +239,130 @@ func (s *Service) processValidators(dutiesResp *ProposerDutiesResponse) (map[uin
 
 	return validators, nil
 }
+
+func (s *Service) scheduleNotificationForSlot(epoch uint64, slot uint64, info *validatorapiv1.SlotInfo) {
+	slotStartTime := s.genesisTime.Add(time.Duration(slot) * SlotDuration)
+	notificationTime := slotStartTime.Add(-NotifyOffset)
```
aloknerurkar, line 245: I dont think we need this delay. We already have the delay to query the duties. I think we can send this as soon as we get the info. In case of the notifications, users will be always subscribed ideally. So sooner they get it, sooner they can decide to bid.

If they miss this, they should be able to query the get validators endpoint right?

p2p/pkg/rpc/validator/service.go
```diff
@@ -29,21 +32,41 @@ type Service struct {
...
+) (*Service, error) {
+	s := &Service{
 		apiURL:          apiURL,
 		validatorRouter: validatorRouter,
 		logger:          logger,
 		metrics:         newMetrics(),
 		optsGetter:      optsGetter,
+		notifier:        notifier,
+	}
+	genesisTime, err := s.fetchGenesisTime(context.Background())
```
mrekucci, line 62: The context should be passed in the `NewService` constructor or at least have a timeout.

p2p/pkg/rpc/validator/service_test.go
```diff
@@ -34,129 +39,356 @@ func (m *MockValidatorRouterContract) AreValidatorsOptedIn(_ *bind.CallOpts, val
...
+
+func (m *MockNotifier) Notify(n *notifications.Notification) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	m.Notifications = append(m.Notifications, n)
+	m.NotifyCh <- n
+}
+
+// --- Override timing globals for tests ---
+func init() {
```
mrekucci, line 63: You should use `export_test.go` file instead of `init()`.

mrekucci (approved):

aloknerurkar (approved):

aloknerurkar (approved):

PR #599, opened 2025-02-13
feat: mainnet first 

ckartik:


tools/points-service/api.go
```diff
@@ -23,15 +23,15 @@ func NewPointsAPI(logger *slog.Logger, db *sql.DB, ps *PointsService) *PointsAPI
 }
 
 func (p *PointsAPI) StartAPIServer(ctx context.Context, addr string) error {
-	r := http.NewServeMux()
-	r.HandleFunc("GET /health", p.HealthCheck)
-	r.HandleFunc("GET /last_block", p.GetLastBlock)
-	r.HandleFunc("GET /{receiver_type}/{receiver_address}", p.RecomputePointsForAddress)
-	r.HandleFunc("GET /stats", p.GetTotalPointsStats)
-	r.HandleFunc("GET /all", p.GetAllPoints)
+	r := mux.NewRouter()
```
mrekucci, line 26: I recommend that you do not make this change on the Gorilla Mux router. We should get rid of it and use the one from the standard library in a separate PR.

tools/points-service/main.go
```diff
@@ -81,7 +80,7 @@ var (
 		Name:    "ethereum-rpc-url",
 		Usage:   "URL of the Ethereum RPC server",
 		EnvVars: []string{"POINTS_ETH_RPC_URL"},
-		Value:   "https://eth-holesky.g.alchemy.com/v2/0DDo7YeieNEucZX3jieFfzmzOCGTKAgp",
+		Value:   "https://eth-mainnet.g.alchemy.com/v2/0DDo7YeieNEucZX3jieFfzmzOCGTKAgp",
```
mrekucci, line 83: The API key should not be here. Instead, there should be some public node as default.

tools/points-service/main.go
```diff
@@ -95,7 +94,14 @@ var (
...
 		EnvVars: []string{"POINTS_MAINNET"},
-		Value:   false,
+		Value:   true,
+	}
+
+	optionStartBlock = &cli.Int64Flag{
+		Name:    "start-block",
+		Usage:   "Block number to start processing from",
+		EnvVars: []string{"POINTS_START_BLOCK"},
+		Value:   21344601,
```
mrekucci, line 104: An explanatory comment should be made as to why this block number was selected.

tools/points-service/api.go
ckartik: The standard library doesn't work properly for me. In the effort of getting this into production on time, I'd say keep the gorilla mux and we can move to the standard library one  after we've launched.

mrekucci (approved):

PR #600, opened 2025-02-13
fix: lower default base and priority fee

Mikelle:
## Describe your changes

Lower default base and priority fee

### DIFF
--- a/p2p/cmd/main.go
+++ b/p2p/cmd/main.go
@@ -376,7 +376,7 @@ var (
 		Name:     "gas-tip-cap",
 		Usage:    "Use predefined gas tip cap for transactions",
 		EnvVars:  []string{"MEV_COMMIT_GAS_TIP_CAP"},
-		Value:    "1000000000", // 1 gWEI
+		Value:    "50000000", // 0.05 gWEI
 		Category: categoryGlobal,
 	})
 
@@ -384,7 +384,7 @@ var (
 		Name:     "gas-fee-cap",
 		Usage:    "Use predefined gas fee cap for transactions",
 		EnvVars:  []string{"MEV_COMMIT_GAS_FEE_CAP"},
-		Value:    "2000000000", // 2 gWEI
+		Value:    "10000", // 0,00001 gWEI
 		Category: categoryGlobal,
 	})
 
--- a/p2p/cmd/main.go
+++ b/p2p/cmd/main.go
@@ -384,7 +384,7 @@ var (
 		Name:     "gas-fee-cap",
 		Usage:    "Use predefined gas fee cap for transactions",
 		EnvVars:  []string{"MEV_COMMIT_GAS_FEE_CAP"},
-		Value:    "10000", // 0,00001 gWEI
+		Value:    "60000000", // 0.06 gWEI
 		Category: categoryGlobal,
 	})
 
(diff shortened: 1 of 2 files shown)


chrmatt (commented):
Numbers look good to me. I cannot judge whether these changes are sufficient overall though. Hence I'm not approving and leave this to somebody with a better understanding of the overall system.

mrekucci (approved):

p2p/cmd/main.go
```diff
@@ -376,15 +376,15 @@ var (
...
+		Value:    "50000000", // 0.05 gWEI
 		Category: categoryGlobal,
 	})
 
 	optionGasFeeCap = altsrc.NewStringFlag(&cli.StringFlag{
 		Name:     "gas-fee-cap",
 		Usage:    "Use predefined gas fee cap for transactions",
 		EnvVars:  []string{"MEV_COMMIT_GAS_FEE_CAP"},
-		Value:    "2000000000", // 2 gWEI
+		Value:    "10000", // 0,00001 gWEI
```
aloknerurkar, line 387: Fee cap is the maxFeePerGas. If you check the generated clients and see how they create the transactions, you will see this calculation:

maxFeePerGas = gasTipCap + (baseFee * 2)

So if our baseFee is 7wei, the maxFeePerGas should be >0.05014.

aloknerurkar: Also, update these values in oracle as well.

Maybe safer to keep gas fee cap as 0.051 or something.

chrmatt: So these values set maxima for the fees? What we've discussed is that we want to change the minimum. For example, we want to ensure that if somebody submits a tx with a fee per gas of 0.04 GWEI, we need to ensure that this transaction cannot make it to the mev-commit chain. If I understand your comment correctly, this is not achieved by this code.

aloknerurkar: So this is purely on client side. For restricting all users, we will need to configure our geth instance to set some values while starting the node. This is not part of the PR. I dont think we can set a minimum, but we can set the max. If user uses lesser values, they might still be able to land their transactions, but it will be lower priority. So they will be late in making it to the chain.

In order to set minimum, we might have to change some things in the geth codebase itself. It can be done but not through configuration.

Mikelle: > I dont think we can set a minimum, but we can set the max. If user uses lesser values, they might still be able to land their transactions, but it will be lower priority. So they will be late in making it to the chain.

> In order to set minimum, we might have to change some things in the geth codebase itself. It can be done but not through configuration.

Actually you can set it with `--txpool.pricelimit=50000000` and `--miner.gasprice=50000000`. If you take a look at the geth code, it will reject txs with less value, than that

Mikelle: > Also, update these values in oracle as well.

Why do we need to update it in the oracle, if oracle txs are free? With higher priority fee we could be sure, that node will select those txs with higher priority

Mikelle: --txpool.pricelimit value           (default: 1)                       ($GETH_TXPOOL_PRICELIMIT)
          Minimum gas price tip to enforce for acceptance into the pool

    --miner.gasprice value              (default: 0)                       ($GETH_MINER_GASPRICE)
          Minimum gas price for mining a transaction

PR #605, opened 2025-02-15
feat: adds manual points alloc endpoint

ckartik:
* Adds the ability to manually add points into the service


mrekucci (approved):

tools/points-service/main.go
```diff
@@ -474,6 +481,37 @@ func insertOptInWithVault(db *sql.DB, logger *slog.Logger, pubkey, adder, vaultA
 	}
 }
 
+func insertManualValRecord(db *sql.DB,
+	logger *slog.Logger,
+	pubkey,
+	adder string,
+	optedInBlock uint64,
+) error {
+	rwLock.RLock()
```
Mikelle, line 490: This lock is not needed, `sql.DB` is already handling concurrent access

tools/points-service/main.go
```diff
@@ -474,6 +481,37 @@ func insertOptInWithVault(db *sql.DB, logger *slog.Logger, pubkey, adder, vaultA
...
+	defer rwLock.RUnlock()
+
+	var existingPubkey string
+	err := db.QueryRow(`
+		SELECT pubkey FROM validator_records
```
Mikelle, line 495: This select is not needed, we can check err after insertion, like this:

`if strings.Contains(err.Error(), "UNIQUE constraint failed") || 
           strings.Contains(err.Error(), "duplicate key value")`

Mikelle (changes requested):
Probably it will be better to address those comments before merging

tools/points-service/main.go
shaspitz: Here I'm trying to enforce uniqueness exclusively for the `pubkey`. The current table has `UNIQUE(pubkey, adder, opted_in_block)` which seems to only enforce uniqueness for a tuple of those values.

Your suggestion is valid, but it'd require adding a new uniqueness constraint to the existing db, which seems like a riskier change than the current naive logic

tools/points-service/main.go
shaspitz: Nice catch, I was pattern matching. Might as well remove the lock altogether then https://github.com/primev/mev-commit/commit/fae763ed7ada7861151c02fc8b24ded8b25eddc3

Mikelle (approved):

shaspitz:
See https://github.com/primev/mev-commit/pull/605/commits/721b1a93d0d22be7928589a0037853b715f8c3e3. After more e2e testing I realized it'd be best to create a new func for manual opt outs which:

1. Enforces the original val record was opted in manually
2. Allows for replacing manual opt-out block as needed

PR #610, opened 2025-02-21
feat: instant bridge service

aloknerurkar:

Fixes # (issue)


tools/instant-bridge/bidder/bidder.go
```diff
@@ -0,0 +1,263 @@
...
+	"sync/atomic"
+	"time"
+
+	bidderapiv1 "github.com/primev/mev-commit/p2p/gen/go/bidderapi/v1"
+	debugapiv1 "github.com/primev/mev-commit/p2p/gen/go/debugapi/v1"
+	notificationsapiv1 "github.com/primev/mev-commit/p2p/gen/go/notificationsapi/v1"
+)
+
+const (
+	epochNotificationTopic = "epoch"
```
Mikelle, line 18: Isn't topic name is `epoch_validators_opted_in`?

p2p/pkg/rpc/validator/service.go
```diff
@@ -273,30 +281,29 @@ func (s *Service) processEpoch(ctx context.Context, epoch uint64, epochTime int6
 
 	optedInSlots := make([]map[string]interface{}, 0)
 	for slot, info := range validators {
+		optedInSlots = append(optedInSlots, map[string]interface{}{
```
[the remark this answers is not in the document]
aloknerurkar: In order to estimate the time we need the exact index of the slot in the epoch. So if we send just the opted in slots, we have to calculate the index of the slot in the epoch which was more complicated.

Other option is to put the start time of each slot in the notification. But I thought this is easier. Wdyt?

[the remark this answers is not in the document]
Mikelle: We could add index/time here in the slot info: https://github.com/primev/mev-commit/blob/main/p2p/pkg/rpc/validator/service.go#L227

tools/instant-bridge/bidder/bidder.go
```diff
@@ -0,0 +1,316 @@
...
+				continue
+			}
+
+			epoch, err := parseEpochInfo(msg)
+			if err != nil {
+				b.logger.Error("failed to parse epoch info", "error", err, "msg", msg)
+				continue
+			}
+
+			b.nextEpoch.Store(epoch)
```
shaspitz, line 121: As mentioned on the call we likely need to check the proposer slot assignments _during the epoch in question_, as proposer selection is only stable within the context of the current epoch. See specification [here](https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/validator.md#validator-assignments) and [here](https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/validator.md#lookahead)

infrastructure/nomad/playbooks/templates/jobs/instant-bridge.nomad.j2
```diff
@@ -0,0 +1,191 @@
...
+            {{- end }}
+            {{- with secret "secret/data/mev-commit" }}
+          ADDRESS="$(cat "${INSTANT_BRIDGE_KEYSTORE_FILE}" | jq -r '.address')"
+            {{ end }}
+          cast send \
+            --keystore "${CONTRACT_DEPLOYER_KEYSTORE_FILE}" \
+            --password "${CONTRACT_DEPLOYER_KEYSTORE_PASSWORD}" \
+            --priority-gas-price 2000000000 \
+            --gas-price 5000000000 \
+            --value 100ether \
```
shaspitz, line 165: Why is 100eth transferred here?

infrastructure/nomad/playbooks/templates/jobs/instant-bridge.nomad.j2
```diff
@@ -0,0 +1,191 @@
...
+          INSTANT_BRIDGE_KEYSTORE_FILENAME="{{ with secret "secret/data/mev-commit" }}{{ .Data.data.instant_bridge_keystore_filename }}{{ end }}"
+          INSTANT_BRIDGE_KEYSTORE_PASSWORD="{{ with secret "secret/data/mev-commit" }}{{ .Data.data.instant_bridge_keystore_password }}{{ end }}"
+          {{- range nomadService "mev-commit-geth-bootnode1" }}
+            {{- if contains "http" .Tags }}
+          INSTANT_BRIDGE_SETTLEMENT_RPC_URL="http://{{ .Address }}:{{ .Port }}"
+            {{- end }}
+          {{- end }}
+          {{- range nomadService "{% endraw %}{{ job.target.name }}{% raw %}" }}
+            {{- if contains "rpc" .Tags }}
+          INSTANT_BRIDGE_BIDDER_RPC_URL="{{ .Address }}:{{ .Port }}"
```
shaspitz, line 96: Does the instant bridge's bidder node run within the instant bridge process or is it a separate process?

Mikelle (approved):

tools/instant-bridge/api/api.go
```diff
@@ -0,0 +1,324 @@
...
+			BidAmountSpent:     bidAmountSpent.String(),
+			FeesAccumulated:    feesAccumulated.String(),
+			L1Balance:          l1Balance.String(),
+			SettlementBalance:  settlementBalance.String(),
+		}); err != nil {
+			a.logger.Error("failed to write response", "error", err)
+		}
+	})
+
+	a.mux.HandleFunc("POST /bid", func(w http.ResponseWriter, req *http.Request) {
```
shaspitz, line 155: This endpoint is called by the bidder cli correct?

infrastructure/nomad/playbooks/templates/jobs/instant-bridge.nomad.j2
aloknerurkar: This is used only when we run mock L1. So it is used to transfer initial funds to the service on L1.

infrastructure/nomad/playbooks/templates/jobs/instant-bridge.nomad.j2
aloknerurkar: Its a separate process. We configure the target here. So the IP/Port of the running bidder node configured. It is configured in profiles.yml as 'target'.

tools/instant-bridge/bidder/bidder.go
aloknerurkar: Updated the notification on mev-commit to only send current epoch information. Also reworded this to current epoch. So now we should only get current epoch related information.

PR #612, opened 2025-02-26
feat: updates api to adhere to symbiotic requirements

ckartik:
* There were some small modifications required to the structure of the points api responses to adhere to symbiotic requirements

### DIFF
--- a/tools/points-service/main.go
+++ b/tools/points-service/main.go
@@ -838,7 +838,6 @@ func main() {
 
 			go func() {
 				for err := range sub.Err() {
-					ps.SetSubscriptionActive(false)
 					logger.Error("subscription error", "error", err)
 				}
 			}()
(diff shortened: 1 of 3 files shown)


tools/points-service/main.go
```diff
@@ -838,7 +838,6 @@ func main() {
 
 			go func() {
 				for err := range sub.Err() {
-					ps.SetSubscriptionActive(false)
```
aloknerurkar, line 841: Why was this removed?

aloknerurkar (approved):

tools/points-service/main.go
ckartik: It was incorrectly setting it to false, there's some common errors that can occur while processing events we don't want procssed, and it shouldn't be setting to inactive, since events are still being subscribed to

PR #613, opened 2025-02-27
feat: update provider registry

Mikelle:
## Describe your changes

Add script to update provider registry contract


shaspitz (commented):
What feat/fix is being introduced in the v2 contract?

Mikelle:
> What feat/fix is being introduced in the v2 contract?

Added support for slash field

Mikelle:
The slash function now accounts for the explicit slashAmt when calculating the amounts to deduct from a provider’s stake.

shaspitz (approved):

PR #635, opened 2025-03-28
fix: fix double notificaions for validator api

Mikelle:
## Describe your changes

Fixed bug with duplicate notifications

### DIFF
--- a/p2p/pkg/rpc/validator/service.go
+++ b/p2p/pkg/rpc/validator/service.go
@@ -35,13 +35,16 @@ type Service struct {
 	notifier             notifications.Notifier
 	genesisTime          time.Time
 	proposerNotifyOffset time.Duration
+
+	slotDuration  time.Duration
+	epochSlots    int
+	epochDuration time.Duration
 }
 
-var (
+const (
 	SlotDuration  = 12 * time.Second
 	EpochSlots    = 32
 	EpochDuration = SlotDuration * time.Duration(EpochSlots)
-	FetchOffset   = 2 * time.Second
 )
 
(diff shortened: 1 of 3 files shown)


p2p/pkg/rpc/validator/service.go
```diff
@@ -328,29 +334,25 @@ func (s *Service) Start(ctx context.Context) <-chan struct{} {
 	eg, egCtx := errgroup.WithContext(ctx)
 
 	eg.Go(func() error {
-		tick := time.NewTicker(EpochDuration / 2)
-		defer tick.Stop()
-
-		currentEpoch := uint64(time.Since(s.genesisTime) / EpochDuration)
+		currentEpoch := uint64(time.Since(s.genesisTime) / s.epochDuration)
```
shaspitz, line 337: AFAIK the slot duration since the merge is exactly 12 seconds, but before the merge it could vary a bit. Would it be better to compute `currentEpoch` here from the merge block (15537393) timestamp?

Mikelle: @shaspitz Genesis time is the time since the merge block, not since the first block.

https://www.quicknode.com/docs/ethereum/eth-v1-beacon-genesis
> The genesis_time configured for the beacon node, which is the unix time in seconds at which the Eth2.0 chain began

shaspitz: Ah perfect 👍

shaspitz (approved):

aloknerurkar (approved):

PR #643, opened 2025-04-01
fix: add withdraw to autodeposit start

Mikelle:
## Describe your changes

Added withdrawal when autodeposit starts


p2p/pkg/autodepositor/autodepositor.go
```diff
@@ -201,28 +240,9 @@ func (adt *AutoDepositTracker) startAutodeposit(egCtx context.Context, eg *errgr
...
-					txn, err := adt.brContract.WithdrawFromWindows(opts, withdrawWindows)
-					if err != nil {
-						return err
-					}
-					adt.logger.Info("withdraw from windows", "hash", txn.Hash(), "windows", withdrawWindows)
-					err = adt.store.ClearDeposits(egCtx, withdrawWindows)
-					if err != nil {
-						return fmt.Errorf("failed to clear deposits: %w", err)
-					}
+				err := adt.withdrawFromPastWindows(egCtx, window.Window)
```
shaspitz, line 243: Isn't `withdrawFromPastWindows` already called in `Start`? What is the scenario where `withdrawFromPastWindows` should be called twice?

Mikelle: Autodeposit is an ongoing process, and once a window has passed, we want to withdraw the funds from that window

shaspitz: Ah my bad makes sense, I thought existing functionality lived elsewhere. All looks good

shaspitz (approved):

PR #650, opened 2025-04-07
fix: added oracle deployment

Mikelle:
## Describe your changes

Added oracle upgrade script


contracts/contracts/core/Oracle.sol
```diff
@@ -14,12 +14,21 @@ import {Errors} from "../utils/Errors.sol";
  * @title Oracle
  * @notice A contract for Fetching L1 Block Builder Info and Block Data.
  * @dev This contract serves as an oracle to fetch and process Ethereum Layer 1 block data.
+ * @custom:oz-upgrades-from Oracle
```
shaspitz: I don't think this line should be in the v1 version of the contract

shaspitz (commented):
I think the v2 should be merged into the release branch

PR #659, opened 2025-04-15
feat: introduce full bidder bot type

shaspitz:
## Describe your changes

Introduces a new mode for the bidder bot where it uses a "full notifier", causing the bot to bid for every block. The old functionality uses a "selective notifier" where bids are only sent for opted-in proposers. 

### DIFF
--- a/tools/bidder-bot/notifier/full.go
+++ b/tools/bidder-bot/notifier/full.go
@@ -1,0 +1,143 @@
+package notifier
+
+import (
+	"context"
+	"log/slog"
+	"sync"
+	"time"
+
+	"github.com/ethereum/go-ethereum"
+	"github.com/ethereum/go-ethereum/core/types"
+)
+
+const (
+	BlockDuration = 12 * time.Second
+)
+
+type FullNotifier struct {
+	logger               *slog.Logger
+	targetBlockNumChan   chan uint64
+	l1Client             L1Client
+	notifySecondsAhead   time.Duration
+	lastNotifiedBlockNum uint64
+	mu                   sync.Mutex
+}
+
+func NewFullNotifier(
+	logger *slog.Logger,
+	targetBlockNumChan chan uint64,
+	l1Client L1Client,
+	notifySecondsAhead time.Duration,
+) *FullNotifier {
+	return &FullNotifier{
+		logger:             logger,
+		targetBlockNumChan: targetBlockNumChan,
+		l1Client:           l1Client,
+		notifySecondsAhead: notifySecondsAhead,
+	}
+}
+
+type L1Client interface {
+	SubscribeNewHead(ctx context.Context, ch chan<- *types.Header) (ethereum.Subscription, error)
+}
+
+func (b *FullNotifier) Start(ctx context.Context) <-chan struct{} {
+	done := make(chan struct{})
+	go func() {
+		defer close(done)
+
+		headers := make(chan *types.Header)
+		sub, err := b.l1Client.SubscribeNewHead(ctx, headers)
+		if err != nil {
+			b.logger.Error("failed to subscribe to new heads", "error", err)
+			return
+		}
+
+		b.logger.Info("subscribed to new block headers")
+
+		for {
+			select {
+			case <-ctx.Done():
+				b.logger.Info("context done")
+				return
+
+			case err := <-sub.Err():
+				b.logger.Error("subscription error", "error", err)
+				sub, err = b.l1Client.SubscribeNewHead(ctx, headers)
+				if err != nil {
+					b.logger.Error("failed to resubscribe to new heads", "error", err)
+					return
+				}
+
+			case header := <-headers:
+				if err := b.handleHeader(ctx, header); err != nil {
+					b.logger.Error("error handling header", "error", err)
+				}
+			}
+		}
+	}()
+	return done
+}
+
+func (b *FullNotifier) handleHeader(ctx context.Context, header *types.Header) error {
+	currentBlockNum := header.Number.Uint64()
+	nextBlockNum := currentBlockNum + 1
+
+	now := time.Now()
+	nextBlockTime := now.Add(BlockDuration)
+	notificationTime := nextBlockTime.Add(-b.notifySecondsAhead)
+
+	if notificationTime.Before(now) {
+		b.sendTargetBlockNotification(nextBlockNum)
+		return nil
+	}
+
+	delay := notificationTime.Sub(now)
+
+	b.logger.Debug("scheduling notification",
+		"target_block_number", nextBlockNum,
+		"delay_seconds", delay.Seconds(),
+		"expected_block_time", nextBlockTime)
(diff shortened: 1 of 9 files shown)


tools/bidder-bot/notifier/full.go
```diff
@@ -0,0 +1,146 @@
...
+			}
+		}
+	}()
+}
+
+func (b *FullNotifier) sendTargetBlockNotification(targetBlockNum uint64) error {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+
+	if targetBlockNum <= b.lastNotifiedBlockNum {
```
aloknerurkar, line 127: This check could be done way before. Right after we get the header.

tools/bidder-bot/notifier/full.go
```diff
@@ -0,0 +1,146 @@
...
+				if err := b.HandleHeader(ctx, header); err != nil {
+					b.logger.Error("error handling header", "error", err)
+				}
+			}
+		}
+	}()
+	return done
+}
+
+func (b *FullNotifier) HandleHeader(ctx context.Context, header *types.Header) error {
```
aloknerurkar, line 83: I dont think these two functions are needed. We should send the target block number notification immediately. If its the next block there is no need to wait as providers will accept bids.

aloknerurkar, line 83: Why is this function exported?

tools/bidder-bot/notifier/full.go
```diff
@@ -0,0 +1,146 @@
...
+	b.logger.Debug("scheduling notification",
+		"target_block_number", targetBlockNum,
+		"notification_time", notificationTime,
+		"expected_block_time", nextBlockTime)
+
+	b.scheduleNotification(ctx, targetBlockNum, notificationTime)
+	return nil
+}
+
+func (b *FullNotifier) scheduleNotification(ctx context.Context, targetBlockNum uint64, notificationTime time.Time) {
```
aloknerurkar, line 105: There is again no need to schedule notification. You can just wait till time passes as your consumer is not going to get more blocks till time passes. The extra go  routine is not required.

tools/bidder-bot/notifier/full.go
shaspitz: Just to test. Changed to export pattern in https://github.com/primev/mev-commit/commit/71ae460bfc9edc10f162526a04b370a39098f5db

shaspitz: `handleHeader` is its own function s.t. it can be tested separately from logic in `Start`. Will reply separately to your comment on `scheduleNotification`

tools/bidder-bot/notifier/full.go
shaspitz: Sending the bid at a specific time during the block sealing process was a request from @iowar. He requested that bids are sent ~7 seconds before the block is sealed if I remember correctly. This is why we delay the notification. To remove this functionality, I'd first need confirmation from Iowar that it would be ok

tools/bidder-bot/notifier/full.go
shaspitz: Intention here is to have this read be a part of the atomic operation enforced by the mutex. Otherwise a race condition is possible where a duplicate notification is sent. 

However if you think I'm being too defensive here, and we should trust `SubscribeNewHead` will not provide duplicates, I can remove `lastNotifiedBlockNum` and the mutex altogether

tools/bidder-bot/notifier/full.go
shaspitz: Simplified things in https://github.com/primev/mev-commit/commit/5ce105a78e93e2e88482b350621249a05573322c

tools/bidder-bot/notifier/full.go
```diff
@@ -0,0 +1,106 @@
...
+	b.logger.Debug("handling header", "target_block_number", targetBlockNum)
+
+	b.mu.Lock()
+	defer b.mu.Unlock()
+
+	if targetBlockNum <= b.lastNotifiedBlockNum {
+		return fmt.Errorf("skipping notification for duplicate target block number %d", targetBlockNum)
+	}
+
+	select {
```
aloknerurkar, line 91: This pattern is problematic. If the consumer is slow this might seem like it would do the job, but if the consumer is blocked, this wouldnt handle the case. It might be better to just have a context.WithTimeout here and log error if this was not able to perform the operation in a sane interval.

tools/bidder-bot/notifier/full.go
```diff
@@ -0,0 +1,106 @@
...
+	case b.targetBlockNumChan <- targetBlockNum:
+		b.logger.Debug("sent target block number", "target_block_number", targetBlockNum)
+	default:
+		select {
+		case drainedTargetBlockNum := <-b.targetBlockNumChan:
+			b.logger.Warn("drained buffered target block number", "drained_target_block_number", drainedTargetBlockNum)
+		default:
+		}
+		b.targetBlockNumChan <- targetBlockNum
```
aloknerurkar, line 100: This is not the case. If we hit the default case here, we dont drain the buffer and are again blocked here. So this logic doesnt make sense.

tools/bidder-bot/notifier/selective.go
```diff
@@ -84,17 +87,31 @@ func (b *Notifier) handleMsg(msg *notificationsapiv1.Notification) error {
...
+		return err
+	}
+
+	// Assume the two slots before upcoming proposer slot are NOT missed
+	targetBlockNum := upcomingSlotMinusTwoBlockNum + 2
+
 	select {
-	case b.proposerChan <- upcomingProposer:
-		b.logger.Debug("sent upcoming proposer", "proposer", upcomingProposer)
+	case b.targetBlockNumChan <- targetBlockNum:
```
aloknerurkar, line 105: Same comment as the full notifier.

tools/bidder-bot/notifier/full.go
shaspitz: There's only one thread that writes to and drains the channel, but you're right that the context timeout is easier to follow. Addressed in https://github.com/primev/mev-commit/pull/659/commits/b4588ea8d6acd705af02e09321dd75d3d9c2f330

tools/bidder-bot/notifier/full.go
shaspitz: context with timeout is a simpler solution here 👍 https://github.com/primev/mev-commit/pull/659/commits/b4588ea8d6acd705af02e09321dd75d3d9c2f330

shaspitz:
From e2e testing, I added this fix https://github.com/primev/mev-commit/pull/659/commits/0ea1209c4bfd815781d35d9bdecfb6b3e6a94a07. Now a `TargetBlock` struct is passed through the channel (with block num and timestamp included), s.t. bids can now timeout at the target block's timestamp

aloknerurkar (approved):

PR #660, opened 2025-04-16
fix: removed return for notification

Mikelle:
## Describe your changes

Removed return, as service is failing to send notification for the first slot in epoch


aloknerurkar (approved):

shaspitz:
This PR should be unnecessary after #661

PR #662, opened 2025-04-17
fix: test for processEpoch func

shaspitz:
## Describe your changes

* `TestProcessEpoch` now properly tests that expected notifications are firing, s.t. its a regression test for https://github.com/primev/mev-commit/pull/661
* It now tests for two opted in slot notifs, making it a better regression test for #661, which handles the first slot of an epoch in a special way

### DIFF
--- a/p2p/pkg/rpc/validator/service_test.go
+++ b/p2p/pkg/rpc/validator/service_test.go
@@ -247,9 +247,12 @@ func TestScheduleNotificationForSlot(t *testing.T) {
 }
 
 func TestProcessEpoch(t *testing.T) {
+	t.Parallel()
+
 	dutiesJSON := `{"data":[
         {"pubkey":"0x1234567890abcdef","slot":"1"},
-        {"pubkey":"0xfedcba0987654321","slot":"2"}
+		{"pubkey":"0x7777777777777777","slot":"2"},
+        {"pubkey":"0xfedcba0987654321","slot":"3"}
     ]}`
 
 	mux := http.NewServeMux()
@@ -269,6 +272,9 @@ func TestProcessEpoch(t *testing.T) {
 	mockValidatorRouter := &MockValidatorRouterContract{
 		ExpectedCalls: map[string]interface{}{
 			string(hexutil.MustDecode("0x1234567890abcdef")): validatoroptinrouter.IValidatorOptInRouterOptInStatus{
+				IsVanillaOptedIn: true,
+			},
+			string(hexutil.MustDecode("0x7777777777777777")): validatoroptinrouter.IValidatorOptInRouterOptInStatus{
 				IsVanillaOptedIn: true,
 			},
 			string(hexutil.MustDecode("0xfedcba0987654321")): validatoroptinrouter.IValidatorOptInRouterOptInStatus{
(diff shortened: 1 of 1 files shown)


p2p/pkg/rpc/validator/service_test.go
```diff
@@ -286,47 +286,58 @@ func TestProcessEpoch(t *testing.T) {
 	slotDuration := 12 * time.Second
 	svc := validatorapi.NewService(ts.URL, mockValidatorRouter, logger, optsGetter, mockNotifier, notifyOffset)
 
-	svc.SetGenesisTime(time.Now().Add(100*time.Millisecond - slotDuration))
```
shaspitz, line 289: lets say current time=20s. Then before this PR, genesis_time=8.1s. Upcoming epoch time would be 20.1s, meaning notification time would be 19.1s. Due to this, we were hitting the "notification time already passed for slot", and the notification wasn't firing. 

With new value, genesis_t would be =9.5s, and notification time=20.5s. This lands us in the middle of the 1s notification timeout

p2p/pkg/rpc/validator/service_test.go
```diff
@@ -286,47 +286,58 @@ func TestProcessEpoch(t *testing.T) {
...
-			}
-			if len(slotsSlice) != 1 {
-				t.Fatalf("expected 1 slot, got %d", len(slotsSlice))
+	numValidatorOptedInNotifs := 0
+	numEpochValidatorsOptedInNotifs := 0
+
+	timeout := time.After(1 * time.Second)
+
+	// expect 1 validator opted in and 1 epoch notif
+	for numValidatorOptedInNotifs < 1 || numEpochValidatorsOptedInNotifs < 1 {
```
shaspitz, line 299: Previously we didn't have a for-loop here, so the timeout wasn't working, and the test passed without a `TopicValidatorOptedIn` notification

p2p/pkg/rpc/validator/service_test.go
```diff
@@ -286,47 +292,58 @@ func TestProcessEpoch(t *testing.T) {
...
-			}
-			if len(slotsSlice) != 1 {
-				t.Fatalf("expected 1 slot, got %d", len(slotsSlice))
+	numValidatorOptedInNotifs := 0
+	numEpochValidatorsOptedInNotifs := 0
+
+	timeout := time.After(1*time.Second + slotDuration)
+
+	// expect 2 validator opted in and 1 epoch notif
+	for numValidatorOptedInNotifs < 2 || numEpochValidatorsOptedInNotifs < 1 {
```
shaspitz, line 305: Previously we didn't have a for-loop here, so only the first TopicEpochValidatorsOptedIn notification was checked, and the test passed without any TopicValidatorOptedIn notifications

p2p/pkg/rpc/validator/service_test.go
```diff
@@ -247,17 +247,15 @@ func TestScheduleNotificationForSlot(t *testing.T) {
...
+
 	dutiesJSON := `{"data":[
         {"pubkey":"0x1234567890abcdef","slot":"1"},
-        {"pubkey":"0xfedcba0987654321","slot":"2"}
+		{"pubkey":"0x7777777777777777","slot":"2"},
+        {"pubkey":"0xfedcba0987654321","slot":"3"}
     ]}`
 
 	mux := http.NewServeMux()
-	// Handle genesis requests.
```
shaspitz, line 256: this mock wasn't needed

Mikelle (approved):

PR #673, opened 2025-04-29
fix: changed decay logic

Mikelle:
## Describe your changes

If a provider commits to a bid before the decay start time, it counts as fully decayed. It should instead not decay at all in this case. It is also realistic that providers may commit before, especially since we describe in the docs as a good practice for bidders to add ~300 ms to the current time to the start to not have any decay if providers commit "immediately".

### DIFF
--- a/oracle/pkg/updater/updater_test.go
+++ b/oracle/pkg/updater/updater_test.go
@@ -1322,14 +1322,14 @@ func TestComputeResidualAfterDecay(t *testing.T) {
 			start:  1000,
 			end:    2000,
 			commit: 500,
-			want:   big.NewInt(updater.ONE_HUNDRED_PERCENT),
+			want:   updater.BigOneHundredPercent,
 		},
 		{
 			name:   "Commit At Start",
 			start:  1000,
 			end:    2000,
 			commit: 1000,
-			want:   big.NewInt(updater.ONE_HUNDRED_PERCENT),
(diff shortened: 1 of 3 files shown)


oracle/pkg/updater/updater_test.go
```diff
@@ -1292,6 +1292,175 @@ func TestUpdaterIgnoreCommitments(t *testing.T) {
...
+			want: big.NewInt(50 * updater.PRECISION),
+		},
+	}
+
+	for _, tc := range tests {
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+			got := u.ComputeResidualAfterDecay(tc.start, tc.end, tc.commit)
+			if tc.name == "Commit Very Close To End" {
```
Mikelle, line 1449: @aloknerurkar not sure - why does the computeResidualAfterDecay function use float64 instead of big.Int? Is it because float64 is simpler or faster? If there’s no strong reason, I could refactor the decay calculation to use big.Int.

Mikelle: Otherwise, with current implementation, this test gives flaky result

aloknerurkar: We can use big.Int no issues.

aloknerurkar (approved):

PR #674, opened 2025-05-01
chore: config for releasing points service bin

shaspitz:
## Describe your changes

Gh releaser workflow didn't previously release go binaries for the points service


shaspitz:
Tested locally by running `goreleaser release --config=./tools/points-service/.goreleaser.yml --snapshot --clean` from the top level dir of this repo. I don't have privileges to do so, but invoking [this workflow](https://github.com/primev/mev-commit/actions/workflows/releaser.yml) would fully test this PR

shaspitz:
> This is not really required. The release only includes the binaries that are exported to public. The nomad workflows determine the binaries needed for deployment. https://github.com/primev/mev-commit/blob/main/infrastructure/nomad/playbooks/variables/profiles.yml#L36

The link you've sent would only apply if we re-deploy the entire cluster right? I'm looking to redeploy just the points service

PR #675, opened 2025-05-04
feat: bidder bot retries once when slot is missed 

shaspitz:
## Describe your changes

We've been seeing occasional instances of this error `failed to get block number for upcoming proposer slot - 2` for the prod bidder bot. The cause is when the slot 2 before an opted-in slot is missed, causing the `/eth/v2/beacon/blocks` to return a 404 error. This PR allows the bidder bot to handle that error. In the case of a missed slot 2 before the target slot, the bidder bot retries a single time to query the slot 3 before the target, and accounts for blockNum not incrementing during the missed slot. 


tools/bidder-bot/notifier/selective.go
```diff
@@ -92,17 +96,21 @@ func (b *SelectiveNotifier) handleMsg(ctx context.Context, msg *notificationsapi
...
+			b.logger.Error("failed to get block number for upcoming proposer slot - 3. No more retries", "error", err)
+			return err
+		}
 	}
-
-	// Assume the two slots before upcoming proposer slot are NOT missed
 	targetBlock := bidder.TargetBlock{
-		Num:  mTwoBlocknum + 2,
-		Time: time.Unix(int64(mTwoTimestamp), 0).Add(2 * slotDuration),
+		Num:  blockNumBeforeTarget + 2, // Same handling for either value of slotBeforeTarget
```
shaspitz, line 112: if `slotBeforeTarget` is `upcomingProposer.Slot - 3`, we should still only add 2 here, as one of the slots was missed, causing block num not to be incremented that slot

tools/bidder-bot/notifier/selective.go
```diff
@@ -92,17 +96,21 @@ func (b *SelectiveNotifier) handleMsg(ctx context.Context, msg *notificationsapi
...
+			b.logger.Error("failed to get block number for upcoming proposer slot - 3. No more retries", "error", err)
+			return err
+		}
 	}
-
-	// Assume the two slots before upcoming proposer slot are NOT missed
 	targetBlock := bidder.TargetBlock{
-		Num:  mTwoBlocknum + 2,
-		Time: time.Unix(int64(mTwoTimestamp), 0).Add(2 * slotDuration),
+		Num:  blockNumBeforeTarget + 2, // Same handling for either value of slotTargetOffset
```
shaspitz, line 112: Even if slotTargetOffset is 3, we should still only add 2 here, as one of the slots was missed, causing block num not to be incremented for the missed slot.

For target block time, we should consider the value of slotTargetOffset

tools/bidder-bot/notifier/selective.go
```diff
@@ -92,17 +96,21 @@ func (b *SelectiveNotifier) handleMsg(ctx context.Context, msg *notificationsapi
...
-	mTwoBlocknum, mTwoTimestamp, err := b.beaconClient.getPayloadDataForSlot(timeoutCtx, upcomingSlotMinusTwo)
+	slotTargetOffset := uint64(2)
+	blockNumBeforeTarget, timestampBeforeTarget, err := b.beaconClient.GetPayloadDataForSlot(timeoutCtx, upcomingProposer.Slot-slotTargetOffset)
 	if err != nil {
-		b.logger.Error("failed to get block number for upcoming proposer slot - 2", "error", err)
-		return err
+		b.logger.Warn("failed to get block number for upcoming proposer slot - 2. This likely indicates a missed slot", "error", err)
+		b.logger.Info("retrying with upcoming proposer slot - 3")
+		slotTargetOffset = uint64(3)
+		blockNumBeforeTarget, timestampBeforeTarget, err = b.beaconClient.GetPayloadDataForSlot(timeoutCtx, upcomingProposer.Slot-slotTargetOffset)
```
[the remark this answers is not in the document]
shaspitz: Sounds good, addressed in https://github.com/primev/mev-commit/commit/25b2e9ebe3b9e4bee3ea55d73ada7544df8e7380. Note I lowered timeout s.t. total is less than 6 seconds. Bidder bot is currently notified 7-8 seconds before the upcoming slot starts

tools/bidder-bot/notifier/selective.go
```diff
@@ -90,19 +94,25 @@ func (b *SelectiveNotifier) handleMsg(ctx context.Context, msg *notificationsapi
 	}
 
 	// Upcoming proposer slot hasn't started yet, so query block number for upcoming proposer slot - 2
-	timeoutCtx, cancel := context.WithTimeout(ctx, 5*time.Second)
+	timeoutCtx, cancel := context.WithTimeout(ctx, 3*time.Second)
```
Mikelle, line 97: Ideally, it would be best to create a constant or configuration parameter to avoid magic numbers in the code (not required though)

Mikelle (approved):

tools/bidder-bot/notifier/selective.go
shaspitz: Makes sense, in this specific scenario with rpc timeout the value is semi arbitrary, but agreed with values that have a lot of intention

PR #680, opened 2025-05-06
fix: fixed solhint

Mikelle:
## Describe your changes

New version of solhint got released (https://github.com/protofire/solhint/releases/tag/v5.1.0) and ci is no longer passing.

So this PR changes order of the functions in our smart contracts

### DIFF
--- a/contracts/.solhint.json
+++ b/contracts/.solhint.json
@@ -15,8 +15,6 @@
     "private-vars-leading-underscore": "error",
     "gas-length-in-loops": "error",
     "const-name-snakecase": "error",
-    "contract-name-camelcase": "error",
-    "event-name-camelcase": "error",
     "func-name-mixedcase": "error",
     "immutable-vars-naming": "error",
     "use-forbidden-name": "error",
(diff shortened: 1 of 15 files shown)


contracts/.solhint.json
```diff
@@ -15,8 +15,6 @@
     "private-vars-leading-underscore": "error",
     "gas-length-in-loops": "error",
     "const-name-snakecase": "error",
-    "contract-name-camelcase": "error",
-    "event-name-camelcase": "error",
```
Mikelle, line 19: Those are no longer exist

contracts/.solhint.json
```diff
@@ -25,6 +23,7 @@
     "visibility-modifier-order": "error",
     "reason-string": "error",
     "gas-custom-errors": "off",
-    "max-states-count": ["error", 20]
+    "max-states-count": ["error", 20],
+    "import-path-check": "off"
```
Mikelle, line 27: Not needed for our path imports, for example import like this is not passing lint: 
`
import {Ownable2StepUpgradeable} from "@openzeppelin/contracts-upgradeable/access/Ownable2StepUpgradeable.sol";
`

aloknerurkar (approved):

PR #683, opened 2025-05-09
feat: changed geth library

Mikelle:
## Describe your changes

* Updating geth library
* Moved to the new go linter version


bridge/standard/cmd/relayer/main.go
```diff
@@ -186,14 +186,17 @@ func main() {
 	signal.Notify(sigc, syscall.SIGINT, syscall.SIGTERM)
 	go func() {
 		<-sigc
+		//nolint:errcheck
```
shaspitz, line 189: There's enough `//nolint:errcheck` in this PR that we should disable that check imo

Mikelle: we don't have configuration for linter: .golangci.yml

shaspitz: Can we create one? I don't see the use of adding tens to hundreds of lint exceptions in a PR

bridge/standard/go.mod
```diff
@@ -1,48 +1,49 @@
 module github.com/primev/mev-commit/bridge/standard
 
-go 1.23
+go 1.23.0
+
+toolchain go1.24.0
```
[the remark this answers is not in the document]
Mikelle: it was done automatically after geth lib update

bridge/standard/cmd/relayer/main.go
Mikelle: @aloknerurkar added placeholders

aloknerurkar (approved):

PR #697, opened 2025-06-12
chore: rm pauseable init change

shaspitz:
https://github.com/primev/mev-commit/pull/584 introduced a change to the vanilla registry without following proper guidelines for contract upgrades. That change is reverted in this PR and will be included in a proper contract upgrade PR


shaspitz:
@aloknerurkar tagging you on this PR fyi but I'm gonna merge it

PR #710, opened 2025-06-27
feat: restructure RPC service

aloknerurkar:
## Describe your changes
- Async transaction handling
- Postgres Database integration instead of embedded DB
- Add more RPC APIs
- Added better unit tests for the transaction sending flow.

Fixes # (issue)


tools/preconf-rpc/sender/sender.go
```diff
@@ -0,0 +1,496 @@
...
+					"bidAmount", result.bidAmount.String(),
+				)
+				break BID_LOOP
+			}
+		default:
+		}
+
+		// Wait for block number to be updated to confirm transaction. If failed
+		// we will retry the bid process till user cancels the operation
+		included, err := t.blockTracker.CheckTxnInclusion(ctx, txn.Hash(), result.blockNumber)
```
shaspitz, line 319: Seems like implementation of `blockTracker.CheckTxnInclusion` could be simplified. Instead of waiting for a block to be sealed, calling `BlockByNumber`, then iterating through every tx, we could just check `eth_getTransactionReceipt`

tools/preconf-rpc/handlers/handlers.go
```diff
@@ -32,158 +25,190 @@ var (
...
-		blockNumber int64,
-	) ([]*types.Transaction, error)
-	DeductBalance(ctx context.Context, account common.Address, amount *big.Int) error
-	HasBalance(ctx context.Context, account common.Address, amount *big.Int) bool
+	GetTransactionByHash(ctx context.Context, txnHash common.Hash) (*sender.Transaction, error)
+	GetTransactionsForBlock(ctx context.Context, blockNumber int64) ([]*sender.Transaction, error)
+	GetTransactionCommitments(ctx context.Context, txnHash common.Hash) ([]*bidderapiv1.Commitment, error)
 	GetBalance(ctx context.Context, account common.Address) (*big.Int, error)
-	AddBalance(ctx context.Context, account common.Address, amount *big.Int) error
+	GetCurrentNonce(ctx context.Context, account common.Address) uint64
```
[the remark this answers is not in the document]
aloknerurkar: This will only return cached nonce if there are queued transactions for the user. If there are no queued transactions it proxies the RPC to the underlying L1 RPC.

tools/preconf-rpc/sender/sender.go
aloknerurkar: This function is called only in the cases when we dont have enough commitments (if all providers didnt commit for opted in block) or if it is not opted in block and we got some commitments. The function essentially checks only the block which we bid on to see if we can find the transaction. getTransactionReceipt doesnt have block number param afaik. This is still a stand-in. There should potentially be more checks here, for eg querying relay to see if one of the providers that gave us the commitment actually built the block. For now its simply checks for block transactions.

shaspitz (approved):

PR #714, opened 2025-07-02
feat: feature/add multiple webhooks

kant777:
Adding support for multiple webhooks. updated the tests

### DIFF
--- a/tools/bidder-bot/main.go
+++ b/tools/bidder-bot/main.go
@@ -87,6 +87,13 @@ var (
 		Value:   false,
 	}
 
+	optionBlockInterval = &cli.Uint64Flag{
+		Name:    "block-interval",
+		Usage:   "Only used for full notifier. Headers that are not a multiple of this interval will be skipped",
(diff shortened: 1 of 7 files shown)


tools/validators-monitor/notification/notifier.go
```diff
@@ -0,0 +1,204 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to marshal message: %w", err)
+	}
+
+	var errorOccurred bool
+	for _, webhookURL := range n.webhookURLs {
+		req, err := http.NewRequestWithContext(ctx, "POST", webhookURL, bytes.NewBuffer(messageJSON))
+		if err != nil {
+			n.logger.Error("Failed to create request", "webhook", webhookURL, "error", err)
+			errorOccurred = true
```
aloknerurkar, line 87: Instead of setting this you can use errors.Join to collect all errors and then print them out for better information on failures.

aloknerurkar (commented):
CI is failing

PR #717, opened 2025-07-07
feat: new bidder emulator

aloknerurkar:

Fixes # (issue)

### DIFF
--- a/tools/bidder-emulator/main.go
+++ b/tools/bidder-emulator/main.go
@@ -1,0 +1,263 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"log/slog"
+	"math/big"
+	"math/rand/v2"
+	"os"
+	"os/signal"
+	"path"
+	"slices"
+	"strings"
+	"syscall"
+	"time"
+
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
+	"github.com/ethereum/go-ethereum/core/types"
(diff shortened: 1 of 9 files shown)


tools/bidder-emulator/main.go
```diff
@@ -0,0 +1,271 @@
...
+		})
+
+		from := txtors[0]
+		to := txtors[1]
+
+		// random amount between 0 and 1_000_000_000_000
+		amount := big.NewInt(rand.Int64N(1_000_000_000_000) + 1)
+
+		mtx.Lock(from.Address().Hex())
+		txn, err := from.SendTransaction(ctx, to.Address(), amount)
```
shaspitz, line 227: Shouldn't we not need to separately `SendTransaction` if the tx payload is a part of the bid?

harshsingh1002 (approved):

tools/bidder-emulator/main.go
aloknerurkar: Basically the problem this emulator solves is that we have 2 different services right now. One is the l1-transaction-emulator which creates the L1 transactions and then the bidder-emulator which picks transactions from the blocks to bid on.

The new emulator basically combines the two. This results in better control when we have to scale up. Each worker will first create a transaction on mock L1 and then once the transaction is mined, it will send a bid for the same transaction. Problem is we dont have block-builders, which is why we have to wait for the txn to be mined to know block number to use in the bid. This will enable us to test the entire flow.

PR #739, opened 2025-07-24
fix: remove tech debt in solidity tests 

shaspitz:

Fixes # (issue)


contracts/test/core/ProviderRegistryTest.sol
```diff
@@ -134,19 +134,10 @@ contract ProviderRegistryTest is Test {
...
 
-    function testFail_ProviderStakeAndRegisterMinStake() public {
+    function test_RevertWhen_ProviderStakeAndRegisterMinStake() public {
         vm.deal(provider, 3 ether);
         vm.prank(provider);
-        vm.expectRevert(bytes(""));
-        providerRegistry.registerAndStake{value: 1 wei}();
-    }
-
-    function testFail_ProviderStakeAndRegisterInvalidBLSKey() public {
```
shaspitz, line 144: This test was outdated. `registerAndStake` does not accept bls pubkeys as arguments since https://github.com/primev/mev-commit/pull/523

PR #742, opened 2025-07-25
fix: tuning bidding logic

aloknerurkar:
## Describe your changes
This is a small fix to estimate the opt in slot correctly. Previously we were not considering the time since the last block in the estimate.

Also in the bidder, we have increased the decay start timestamp delay to 200ms. This is more generous and gives providers some leeway to avoid decay. Also increased the decay time to twice the timeout. Which would make the bid decay atmost to half of its value if provider takes the entire timeout duration to respond. This is still crude but need to give more thought on pricing. Especially since we also need to consider clients which are setting the priority fee on their own.

Fixes # (issue)

### DIFF
--- a/tools/preconf-rpc/sender/sender.go
+++ b/tools/preconf-rpc/sender/sender.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
+	"math"
 	"math/big"
 	"strings"
 	"sync"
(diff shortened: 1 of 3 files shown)


tools/preconf-rpc/sender/sender.go
```diff
@@ -528,7 +521,13 @@ func (t *TxSender) sendBid(
 		}
 	}
 
-	cost, blockNo, err := t.calculatePriceForNextBlock(txn, prices)
+	start := time.Now()
+	optedInSlot := timeToOptIn <= blockTime-(prices.MsSinceLastBlock/1000)
```
chrmatt, line 525: Can't we directly check for the slot number? This seems a bit fragile with preicse ms timing etc.

chrmatt (approved):

PR #755, opened 2025-08-04
feat: simplify single node cl implementation 

shaspitz:
The current implementation of our consensus layer is exhibiting odd behavior where recently built block timestamps will jump up to 20 minutes ahead of system time. This PR streamlines implementation, removes unnecessary variables, and makes the logic easier to comprehend/debug. After testing, these changes have fixed the issue. 

* `bb.lastBlockTime` variable was being set independently from `bb.executionHead.BlockTime`. We now use existing `bb.executionHead.BlockTime` in-place
* `head` variable local to `GetPayload` was being set independently from `bb.executionHead`. We now use existing `bb.executionHead` in-place
* `loadExecutionHead` was removed. We now just set `bb.executionHead` once upon service restart in `setExecutionHeadFromRPC` and subsequently update `bb.executionHead` at the end of every `FinalizeBlock`
* Block timestamp computation was simplified
* `bb.buildDelayMs` removed, `bb.buildDelay` remains

### DIFF
--- a/cl/blockbuilder/blockbuilder.go
+++ b/cl/blockbuilder/blockbuilder.go
@@ -55,10 +55,7 @@ type BlockBuilder struct {
 	buildDelay            time.Duration
 	buildEmptyBlocksDelay time.Duration
 	feeRecipient          common.Address
-	// Execution head of the previous block.
-	// This field should only be set upon restart in setExecutionHeadFromRPC,
-	// OR at the end of FinalizeBlock.
-	executionHead *types.ExecutionHead
+	executionHead         *types.ExecutionHead
 }
 
 func NewBlockBuilder(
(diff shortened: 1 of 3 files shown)


cl/blockbuilder/blockbuilder.go
```diff
@@ -153,50 +165,19 @@ func (bb *BlockBuilder) GetPayload(ctx context.Context) error {
...
-		ts = head.BlockTime + 1 // Subsequent blocks must have a higher timestamp.
+	ts := uint64(time.Now().UnixMilli())
+	if ts <= bb.executionHead.BlockTime {
+		bb.logger.Warn(`Leader: Current timestamp is less than or equal to
+			previous block timestamp. Genesis timestamp could have been set incorrectly`,
+			"current_ts", ts,
+			"prev_head_block_time", bb.executionHead.BlockTime,
+		)
+		ts = bb.executionHead.BlockTime + 1
+		time.Sleep(time.Until(time.UnixMilli(int64(ts))))
```
aloknerurkar, line 176: Better to check for ctx cancellation than just sleeping


```suggestion
		select {
		case <-ctx.Done():
		    // Handle Ctx cancellation
		case <-time.After():
		}
```

PR #763, opened 2025-08-21
Lido registry hoodi

owen-eth:
## Describe your changes
Initial registry for Lido V3 Staking Vaults
-No staking/slashing, only freezing
-Operator must be whitelisted
-Operator specifies their vault and keys to register
-Specified Vault's node operator must match the msg.sender, Vault must also be indexed within Lido's Vault Hub
-Lido V3 is not on mainnet yet, this is just for Hoodi test usage

Potential future design:
-Lido Staking Vaults accept ETH and mint steth to users, we can have our registry accept steth so to allow slashing on our side.

Fixes # (issue)


shaspitz (commented):
General flow looks good 👍

PR #766, opened 2025-08-28
feat: bid expression changes

aloknerurkar:
## Describe your changes
- Changes to enable bid expressions starting with position constraints for transactions in a block.
- Updating oracle to check the expressions to determine reward/slash.
- Protocol bump required

Fixes # (issue)

### DIFF
--- a/p2p/cmd/main.go
+++ b/p2p/cmd/main.go
@@ -298,18 +298,16 @@ var (
 	})
 
 	optionTargetDepositAmount = altsrc.NewStringFlag(&cli.StringFlag{
-		Name:  "target-deposit-amount",
-		Usage: "Target deposit amount that'll be set for every valid provider",
-		// Backwards compatible with prev "auto deposit" feature
-		EnvVars:  []string{"MEV_COMMIT_TARGET_DEPOSIT_AMOUNT", "MEV_COMMIT_AUTODEPOSIT_AMOUNT"},
+		Name:     "target-deposit-amount",
+		Usage:    "Target deposit amount that'll be set for every valid provider",
+		EnvVars:  []string{"MEV_COMMIT_TARGET_DEPOSIT_AMOUNT"},
 		Category: categoryBidder,
 	})
 
 	optionEnableDepositManager = altsrc.NewBoolFlag(&cli.BoolFlag{
-		Name:  "enable-deposit-manager",
-		Usage: "Whether the deposit manager should be enabled",
-		// Backwards compatible with prev "auto deposit" feature
-		EnvVars:  []string{"MEV_COMMIT_ENABLE_DEPOSIT_MANAGER", "MEV_COMMIT_AUTODEPOSIT_ENABLED"},
+		Name:     "enable-deposit-manager",
+		Usage:    "Whether the deposit manager should be enabled",
+		EnvVars:  []string{"MEV_COMMIT_ENABLE_DEPOSIT_MANAGER"},
 		Value:    false,
 		Category: categoryBidder,
 	})
--- a/p2p/pkg/preconfirmation/preconfirmation.go
+++ b/p2p/pkg/preconfirmation/preconfirmation.go
@@ -25,7 +25,7 @@ import (
 
 const (
 	ProtocolName    = "preconfirmation"
-	ProtocolVersion = "3.0.0"
+	ProtocolVersion = "4.0.0"
 )
 
 type Preconfirmation struct {


p2p/rpc/bidderapi/v1/bidderapi.proto
```diff
@@ -296,6 +296,53 @@ message WithdrawFromWindowsResponse {
   repeated WithdrawResponse withdraw_responses = 1;
 };
 
+message PositionConstraint {
```
shaspitz, line 299: Can we put this message in a proto file that both the provider and bidder protos share?

oracle/pkg/store/store.go
```diff
@@ -34,16 +34,8 @@ CREATE TABLE IF NOT EXISTS settlements (
 	nonce BIGINT,
 	settled BOOLEAN,
 	decay_percentage BIGINT,
-	settlement_window BIGINT
-);`
-
-var encryptedCommitmentsTable = `
-CREATE TABLE IF NOT EXISTS encrypted_commitments (
```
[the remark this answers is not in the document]
aloknerurkar: I think this was something that was moved out from the oracle and was required for the dashboard. I checked and we dont use this, so can save storage on the oracle DB.

contracts/test/core/PreconfManagerTest.sol
```diff
@@ -89,9 +90,27 @@ contract PreconfManagerTest is Test {
             hex"aeed5b345d04360c6ad52d4fb4fce32eec8a552f87686afb39ceea04f9fd1a782b180e4eef5e02af77015292840c541e2681c8e165b44be1d8276aba7211bde21b",
             hex"a0b508b09c6942d73b8feb4feb308ea0b753e14e32cff231f75348f25feb07b02b97743c4c9c493825ec6730e2bc24a513d09a2996a20beac2013f134c047cd71b",
             15,
-            zkProof
+            zkProof,
+            hex""
         );
 
+        // _testCommitmentAliceBob = TestCommitment(
```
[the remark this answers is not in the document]
aloknerurkar: @owen-eth We also need to adjust the tests in this file to have some sort of dummy options.

contracts/contracts/core/PreconfManager.sol
```diff
@@ -490,7 +491,8 @@ contract PreconfManager is
                         params.decayEndTimeStamp,
                         params.slashAmt,
                         params.zkProof[2], // _bidderPKx,
-                        params.zkProof[3] // _bidderPKy
+                        params.zkProof[3], // _bidderPKy
+                        keccak256(params.bidOptions)
```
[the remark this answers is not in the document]
chrmatt: Ah, you are right. I missed the "variable length" part.

oracle/pkg/updater/updater.go
```diff
@@ -388,6 +356,42 @@ func (u *Updater) handleOpenedCommitment(
...
+						"blockNumber", update.BlockNumber,
+						"constraint", opt.GetPositionConstraint(),
+					)
+					// Remove the satisfied constraint
+					opts.Options = append(opts.Options[:idx], opts.Options[idx+1:]...)
+					break
+				}
+			}
+		}
+		if i == len(commitmentTxnHashes)-1 && len(opts.Options) > 0 {
```
chrmatt, line 376: Why not do this after the loop?

oracle/pkg/updater/updater.go
```diff
@@ -233,42 +175,65 @@ func (u *Updater) Start(ctx context.Context) <-chan struct{} {
...
+		case bidderapiv1.PositionConstraint_ANCHOR_BOTTOM:
+			if txnDetails.PosInBlock >= (len(txns)*(100-int(c.Value)))/100 {
+				return true
+			}
+		}
+	case bidderapiv1.PositionConstraint_BASIS_GAS_PERCENTILE:
+		gasUsed := gasUsedUntil(txnDetails.PosInBlock, txns)
+		gasPercentile := (gasUsed * 100) / txnDetails.TotalGas
+		txnGasPercentile := (txnDetails.GasUsed * 100) / txnDetails.TotalGas
+		if txnGasPercentile > uint64(c.Value) {
```
chrmatt, line 221: I don't think this is always correct: Assume `c.Value = 20`, `txnGasPercentile = 10`, and the transaction comes after 50% of the total block gas. I'd say it should not count as satisfied because it is entirely outside of the requested range.

oracle/pkg/updater/updater.go
```diff
@@ -233,42 +175,65 @@ func (u *Updater) Start(ctx context.Context) <-chan struct{} {
...
+			// The transaction itself uses more gas than the constraint allows
+			return true
+		}
+		switch c.Anchor {
+		case bidderapiv1.PositionConstraint_ANCHOR_TOP:
+			if gasPercentile <= uint64(c.Value) {
+				return true
+			}
+		case bidderapiv1.PositionConstraint_ANCHOR_BOTTOM:
+			if gasPercentile >= uint64(100-c.Value) {
```
chrmatt, line 231: Maybe here we can say `if gasPercentile + txnGasPercentile >= uint64(100-c.Value) {`. That means "some of the gas of the tx is inside the requested range", consistent with requiring some tx of a bundle to be inside the range.

If we do this, I think the `txnGasPercentile > uint64(c.Value)` check above can be removed (it should work as is for top anchor).

oracle/pkg/updater/updater.go
```diff
@@ -362,10 +311,25 @@ func (u *Updater) handleOpenedCommitment(
 		revertableTxnsMap[txn] = true
 	}
 
+	opts := new(bidderapiv1.BidOptions)
+	positionalConstraintsSatisfied := 0
```
chrmatt, line 315: Is this still used?

chrmatt (approved):
Looks good to me.

PR #767, opened 2025-08-29
fix: provider side deposit cache

shaspitz:
PR into https://github.com/primev/mev-commit/pull/746 addressing https://github.com/primev/mev-commit/pull/746#discussion_r2308679584

### DIFF
--- a/p2p/pkg/depositmanager/deposit.go
+++ b/p2p/pkg/depositmanager/deposit.go
@@ -21,10 +21,10 @@ type BidderRegistryContract interface {
 }
 
 type Store interface {
-	GetBalance(bidder common.Address, provider common.Address) (*big.Int, error)
-	SetBalance(bidder common.Address, provider common.Address, balance *big.Int) error
-	DeleteBalance(bidder common.Address, provider common.Address) error
-	RefundBalanceIfExists(bidder common.Address, provider common.Address, amount *big.Int) error
+	GetBalance(bidder common.Address) (*big.Int, error)
+	SetBalance(bidder common.Address, balance *big.Int) error
+	DeleteBalance(bidder common.Address) error
+	RefundBalanceIfExists(bidder common.Address, amount *big.Int) error
 }
 
 type DepositManager struct {
@@ -151,7 +151,7 @@ func (dm *DepositManager) Start(ctx context.Context) <-chan struct{} {
 					continue
 				}
 
-				if err := dm.store.RefundBalanceIfExists(bidder, dm.thisProviderAddress, bidAmountInt); err != nil {
+				if err := dm.store.RefundBalanceIfExists(bidder, bidAmountInt); err != nil {
 					dm.logger.Error("refunding balance", "error", err)
 					return err
 				}
@@ -162,13 +162,13 @@ func (dm *DepositManager) Start(ctx context.Context) <-chan struct{} {
 					dm.logger.Debug("ignoring deposit event for different provider", "provider", deposit.Provider)
 					continue
 				}
-				currentBalance, err := dm.store.GetBalance(deposit.Bidder, deposit.Provider)
+				currentBalance, err := dm.store.GetBalance(deposit.Bidder)
 				if err != nil {
 					dm.logger.Error("getting balance", "error", err)
 					return err
 				}
 				if currentBalance == nil {
-					if err := dm.store.SetBalance(deposit.Bidder, deposit.Provider, deposit.NewAvailableAmount); err != nil {
+					if err := dm.store.SetBalance(deposit.Bidder, deposit.NewAvailableAmount); err != nil {
 						dm.logger.Error("setting balance", "error", err)
 						return err
 					}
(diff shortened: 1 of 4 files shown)


chrmatt:
> Every commitment (relevant to the provider running the node) gets seen by the `openCommitments` function via the for-loop in that function. For every commitment where `providerAddress != newL1Block.Winner`, the pending refund is applied, since a different provider has won the block. Meaning we should re-credit our cached knowledge of bidder deposit specific to our provider and the relevant bidder.

But the `openCommitments` function is only called if any commitment gets opened, I assume? So what if provider A issues commitments for block N, but provider B does not issue any commitments. Now provider A has some pending refunds. Assume provider B wins the block. Then, provider B does not open any commitments. In that case, the `openCommitments` function would not be called and nothing would be refunded for provider A (even though all commitments for that block could/should be refunded).

p2p/pkg/depositmanager/deposit.go
```diff
@@ -181,38 +182,31 @@ func (dm *DepositManager) Start(ctx context.Context) <-chan struct{} {
 	return doneChan
 }
 
-func (dm *DepositManager) CheckAndDeductDeposit(
+func (dm *DepositManager) CheckDeposit(
```
shaspitz, line 185: Per request deduction doesn't happen here and now happens in tracker.go

p2p/pkg/depositmanager/deposit.go
```diff
@@ -23,7 +23,8 @@ type Store interface {
 	GetBalance(bidder common.Address, provider common.Address) (*big.Int, error)
 	SetBalance(bidder common.Address, provider common.Address, balance *big.Int) error
 	DeleteBalance(bidder common.Address, provider common.Address) error
-	RefundBalanceIfExists(bidder common.Address, provider common.Address, amount *big.Int) error
+	IncreaseBalanceIfExists(bidder common.Address, provider common.Address, amount *big.Int) error
+	DecreaseBalanceIfExists(bidder common.Address, provider common.Address, amount *big.Int) error
 }
 
 type DepositManager struct {
```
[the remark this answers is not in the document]
shaspitz: Made the performance fix to the main PR, not this one. See https://github.com/primev/mev-commit/commit/beacd72ead6fe9bdbb23a632726943058027ce05

p2p/pkg/depositmanager/store/store.go
```diff
@@ -64,7 +64,7 @@ func (s *Store) DeleteBalance(bidder common.Address, provider common.Address) er
 	return s.st.Delete(balanceKey(bidder, provider))
 }
 
-func (s *Store) RefundBalanceIfExists(
+func (s *Store) IncreaseBalanceIfExists(
```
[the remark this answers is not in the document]
shaspitz: Changed back

p2p/pkg/preconfirmation/tracker/tracker.go
```diff
@@ -477,6 +477,12 @@ func (t *Tracker) statusUpdater(
 							"txnHash":          task.commitment.Bid.TxHash,
 							"error":            r.Err.Error(),
 						}
+						if task.commitment.BidderAddress != nil {
```
aloknerurkar, line 480: nit: This check can be avoided as you are already ignoring the event in deposit manager if these are not correctly populated. The double check can be avoided.

p2p/pkg/preconfirmation/tracker/tracker.go
```diff
@@ -538,6 +544,24 @@ func (t *Tracker) openCommitments(
 				"providerAddress", commitment.ProviderAddress,
 				"winner", newL1Block.Winner,
 			)
+			if commitment.BidderAddress == nil {
```
aloknerurkar, line 547: Same as above. Double check can be avoided.

p2p/pkg/depositmanager/store/store_test.go
```diff
@@ -91,8 +91,8 @@ func TestStore_RefundBalanceIfExists(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	refundAmount := big.NewInt(5)
-	err = s.RefundBalanceIfExists(bidder, provider, refundAmount)
+	increaseAmount := big.NewInt(5)
+	err = s.RefundBalanceIfExists(bidder, provider, increaseAmount)
```
aloknerurkar, line 95: The changes look good now. But this DB entry can be modified to not have provider address. It will have the same provider address for all entries. So we can save on the extra storage.

aloknerurkar (approved):
I am approving it. Have some minor comments, so you can address them and merge.

p2p/pkg/preconfirmation/tracker/tracker.go
shaspitz: Also necessary since pointer types, I've made this part consistent with the other section https://github.com/primev/mev-commit/commit/394ad8913307007d197b7fba7f13010fcff047a7

p2p/pkg/depositmanager/store/store_test.go
shaspitz: https://github.com/primev/mev-commit/commit/72e608c5a63dd0d7f1b57c6a84c4bbf227ddf6fc

PR #789, opened 2025-09-17
fix: payload update handling in CL

aloknerurkar:

Fixes # (issue)


cl/singlenode/singlenode.go
```diff
@@ -329,24 +339,22 @@ func (app *SingleNodeApp) produceBlock() error {
 	}
 
 	if app.payloadRepo != nil {
-		payloadInfo := &types.PayloadInfo{
+		// Save payload to repository
+		saveCtx, saveCancel := context.WithTimeout(app.appCtx, 400*time.Millisecond)
```
shaspitz, line 343: Do we intend to double the timeout here?

aloknerurkar: It usually finishes early. Seems safer. I can change it back if you think its not needed.

PR #793, opened 2025-09-18
feat: optional redis config for leader follower setup

shaspitz:
Provides the option for both the leader and follower consensus clients to use redis instead of postgres for reading/writing payloads


aloknerurkar (approved):
LGTM!

PR #794, opened 2025-09-18
feat: cl metrics

shaspitz:
Exposes metrics for leader and follower cl services


cl/singlenode/singlenode.go
```diff
@@ -18,13 +18,26 @@ import (
...
+	"github.com/prometheus/client_golang/prometheus/promhttp"
 )
 
 const (
 	// Stop Function
 	shutdownTimeout = 5 * time.Second
 )
 
+var (
+	snDBDuration = promauto.NewHistogramVec(prometheus.HistogramOpts{
```
aloknerurkar, line 32: The promauto pkg registers the collectors also?

shaspitz: Yes promauto automatically registers the metrics it creates with the default registry, and `promhttp.Handler()` uses the default registry

PR #797, opened 2025-09-19
fix: keep postgres connections alive with heartbeat

kant777:
Harden Postgres connection handling to avoid insert timeouts after idle periods.

Details

Pool settings:

- MaxOpenConns=25, MaxIdleConns=20 to keep a warm stash.

- ConnMaxLifetime=0 disables age-based churn.

- ConnMaxIdleTime=0 prevents idle reaping. Ensures the pool never drops healthy connections.

Heartbeat goroutine:

- Runs db.PingContext every 2 minutes.

- Keeps at least one connection active, preventing k8s/LB/NAT or Postgres from closing all idle sessions.

- Interval chosen < common infra idle timeout (~10m).

Why

Without this, long quiet gaps drained the pool; the next insert had to open a new conn (dial + TLS + auth) inside the 5s insert deadline → context deadline exceeded.
These changes ensure a live connection is always available, so inserts stay within the 5s budget even after idle periods.

### DIFF
--- a/cl/singlenode/payloadstore/postgres.go
+++ b/cl/singlenode/payloadstore/postgres.go
@@ -42,16 +42,6 @@ func NewPostgresRepository(ctx context.Context, dsn string, logger *slog.Logger)
 		return nil, fmt.Errorf("failed to ping postgres: %w", err)
 	}
 
-	go func() {
-		t := time.NewTicker(2 * time.Minute)
-		defer t.Stop()
-		for range t.C {
-			c, ccancel := context.WithTimeout(context.Background(), time.Second)
-			_ = db.PingContext(c)
-			ccancel()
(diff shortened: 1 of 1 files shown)


cl/singlenode/payloadstore/postgres.go
```diff
@@ -41,6 +42,16 @@ func NewPostgresRepository(ctx context.Context, dsn string, logger *slog.Logger)
 		return nil, fmt.Errorf("failed to ping postgres: %w", err)
 	}
 
+	go func() {
```
aloknerurkar, line 45: I dont think this is required. We are constantly writing to postgres every 200ms at the very least. Not sure what this ping will do that a write will not.

PR #798, opened 2025-09-19
feat: new validatoroptinhub contract

owen-eth:
## Describe your changes

New ValidatorOptInHub contract which allows future registry support/removal/updates.
-Backwards compatibility with OptInRouter, where the OptInRouter's vanilla address will be set to the Hub. Hub will then check every registry and return true if validator is registered in any of them. Existing AVS and Middleware addresses will be replaced to return false in the OptInRouter since they will be covered by the Hub check.
-IsValidatorOptedIn() now returns an array of booleans instead of the previous OptInStatus struct.
-areValidatorsOptedIn() now returns an array of boolean arrays instead of an array of OptInStatus structs.
-The new contract also contains additional `isValidatorOptedIn**Any**`, which just returns a single bool to signal whether a key is opted-in in **any** of the registries. The same pattern used in an`areValidatorsOptedIn**Any**` function which returns a list of bools to signal whether each key is opted-in in any of the registries

Fixes # (issue)


contracts/contracts/validator-registry/ValidatorOptInHub.sol
```diff
@@ -0,0 +1,153 @@
...
+        if (address(registries[index]) == registry) {
+            registries[index] = IRegistry(address(0));
+            emit RegistryRemoved(index, registry);
+            return;
+        }
+        revert IndexRegistryMismatch();
+    }
+
+    /// @notice Returns an array of bool lists indicating whether each validator pubkey is opted in to mev-commit.
+    function areValidatorsOptedIn(bytes[] calldata valBLSPubKeys) external view returns (bool[][] memory) {
```
shaspitz, line 80: How does the caller know which registry corresponds to which bool for each pubkey?

contracts/contracts/validator-registry/ValidatorOptInHub.sol
```diff
@@ -0,0 +1,153 @@
...
+        if (address(registries[index]) == oldRegistry) {
+            registries[index] = IRegistry(newRegistry);
+            emit RegistryReplaced(index, oldRegistry, newRegistry);
+            return;
+        }
+        revert IndexRegistryMismatch();
+    }
+
+    /// Pass in index and registry address for safer removal.
+    function removeRegistry(uint256 index, address registry) external onlyOwner {
```
shaspitz, line 68: Is there a way to remove the registry and also shrink the length of the array by 1? Instead of zeroing the address

contracts/l1-deployer-cli.sh
```diff
@@ -267,15 +267,15 @@ get_chain_params() {
 }
 
 check_git_status() {
-    if ! current_tag=$(git describe --tags --exact-match 2>/dev/null); then
-        echo "Error: Current commit is not tagged. Please ensure the commit is tagged before deploying."
-        exit 1
-    fi
+    # if ! current_tag=$(git describe --tags --exact-match 2>/dev/null); then
```
shaspitz, line 270: Do we want to remove these checks forever? seems useful to ensure we deploy mainnet contracts from a tagged release

contracts/scripts/validator-registry/DeployValidatorOptInHub.s.sol
```diff
@@ -8,89 +8,67 @@ pragma solidity 0.8.26;
 import {Script} from "forge-std/Script.sol";
```
shaspitz, line 8: Imo we should create an entirely new deployment script since we are introducing a new contract here. Separately we can create another file for updating the old registry to ref the new one for backwards compatibility

shaspitz (commented):
High level design lgtm 👍

contracts/contracts/validator-registry/ValidatorOptInHub.sol
owen-eth: We can do a swap and pop, but I figured it will be rare to remove, and least intrusive to clients using it since it wont mess up the indexes of the other registries

contracts/l1-deployer-cli.sh
owen-eth: forgot to uncomment this, good eye

contracts/contracts/validator-registry/ValidatorOptInHub.sol
owen-eth: Considered adding a string label for read only purposes, but the caller can also just read the registries array to know the registry address at each index

shaspitz: String label would be nice for UX purposes, but optional until there's a concrete need for it

shaspitz (approved):
LGTM

PR #802, opened 2025-09-24
feat: add builder observer data indexer

rose2221:
# feat(indexer): builder-observer data ingestor

Adds comprehensive blockchain indexing with MEV relay bid data and BeaconChain metadata for builder observer functionality.

## Local Setup (StarRocks in Docker)

```bash
# 1) Start StarRocks all-in-one
docker run -d --name starrocks-local \
  -p 9030:9030 -p 8030:8030 -p 8040:8040 \
  starrocks/allin1-ubuntu:latest

# 2) Open MySQL shell into FE
docker exec -it starrocks-local mysql -h 127.0.0.1 -P 9030 -u root
```

## Schema DDL (Updated for Builder Observer)

```sql
CREATE DATABASE IF NOT EXISTS indexer_db;
USE indexer_db;

-- Relays configuration
CREATE TABLE IF NOT EXISTS relays (
  relay_id BIGINT,
  name VARCHAR(100),
  tag VARCHAR(100), 
  base_url VARCHAR(500),
  is_active TINYINT
) ENGINE=OLAP
DUPLICATE KEY(relay_id)
DISTRIBUTED BY HASH(relay_id) BUCKETS 1
PROPERTIES("replication_num" = "1");

-- Blocks with complete builder observer data
CREATE TABLE IF NOT EXISTS blocks (
  -- Core block data
  slot BIGINT,
  block_number BIGINT,
  timestamp DATETIME,
  proposer_index BIGINT,
  
  -- MEV/Relay data
  winning_relay VARCHAR(100),
  winning_builder_pubkey VARCHAR(200),
  fee_recipient VARCHAR(200),
  producer_reward_eth DECIMAL(20,8),
  
  -- Validator data
  validator_pubkey VARCHAR(200),
  validator_opted_in TINYINT,
  
  -- BeaconChain.in metadata for builder observer
  block_status INT,              -- 0=missed, 1=proposed, 2=orphaned
  graffiti VARCHAR(64),          -- Block graffiti
  block_root VARCHAR(66),        -- Block root hash
  parent_root VARCHAR(66),       -- Parent block root
  state_root VARCHAR(66),        -- State root hash
  
  -- Tracking
  created_at DATETIME DEFAULT CURRENT_TIMESTAMP
) ENGINE=OLAP
DUPLICATE KEY(slot)
DISTRIBUTED BY HASH(slot) BUCKETS 10
PROPERTIES("replication_num" = "1");

-- MEV relay bids
CREATE TABLE IF NOT EXISTS bids (
  slot BIGINT,
  relay_id BIGINT,
  builder_pubkey VARCHAR(200),
  proposer_pubkey VARCHAR(200),
  proposer_fee_recipient VARCHAR(200),
  value_wei VARCHAR(100),
  block_number BIGINT,
  timestamp_ms BIGINT
) ENGINE=OLAP
DUPLICATE KEY(slot, relay_id, builder_pubkey)
DISTRIBUTED BY HASH(slot) BUCKETS 10
PROPERTIES("replication_num" = "1");

-- Application state and configuration storage
CREATE TABLE IF NOT EXISTS state (
  key_name VARCHAR(100) PRIMARY KEY,
  value VARCHAR(1000),
  updated_at DATETIME DEFAULT CURRENT_TIMESTAMP
) ENGINE=OLAP
DUPLICATE KEY(key_name)
DISTRIBUTED BY HASH(key_name) BUCKETS 1
PROPERTIES("replication_num" = "1");

## Environment Variables

Required:
```bash
DATABASE_URL=root@tcp(127.0.0.1:9030)/indexer_db
INFURA_RPC=https://mainnet.infura.io/v3/YOUR_KEY
BEACON_BASE=https://beaconcha.in/api/v1  # For BeaconChain.in API
```

Optional:
```bash
ETHERSCAN_API_KEY=your_key_here
HTTP_TIMEOUT=15s
BLOCK_TICK=12s
VALIDATOR_WAIT=1.5s
BACKFILL_EVERY=5m0s
```

## Export and Run

```bash
export DATABASE_URL=root@tcp(127.0.0.1:9030)/indexer_db
export INFURA_RPC=https://mainnet.infura.io/v3/YOUR_KEY
export BEACON_BASE=https://beaconcha.in/api/v1

go run indexer/cmd/indexer/main.go
```


tools/indexer/pkg/database/starrock.go
```diff
@@ -0,0 +1,193 @@
...
+		return 0, false
+	}
+	return bn, true
+}
+
+func (db *DB) SaveLastBlockNumber(ctx context.Context, bn int64) error {
+	ctx2, cancel := context.WithTimeout(ctx, 3*time.Second)
+	defer cancel()
+
+	_, err := db.Conn.ExecContext(ctx2, `DELETE FROM ingestor_state WHERE id = 1`)
```
aloknerurkar, line 92: Both these operations should be done atomically in a transaction.

tools/indexer/cmd/indexer/main.go
```diff
@@ -0,0 +1,434 @@
...
+	log.Printf("[RELAYS] Loaded %d active relays:", len(relays))
+	for _, r := range relays {
+		log.Printf("  - Relay ID %d: %s", r.ID, r.URL)
+	}
+
+	// Initialize starting block number
+	lastBN, found := db.LoadLastBlockNumber(ctx)
+	if !found || lastBN == 0 {
+		log.Printf(" [INIT] No previous state found, checking database for latest block...")
+		err := db.Conn.QueryRowContext(ctx, `SELECT COALESCE(MAX(block_number),0) FROM blocks`).Scan(&lastBN)
```
aloknerurkar, line 183: If the point of the DB is to mask the underlying storage layer, we shouldnt be accessing the Conn directly. Also if the DB layer is changed, for eg a new table instead of blocks is added, this will also have to be changed. The Conn should be unexported. Any functionality should be provided with the help of functions.

tools/indexer/cmd/indexer/main.go
```diff
@@ -0,0 +1,434 @@
...
+	infuraRPC := c.String(optionInfuraRPC.Name)
+	beaconBase := c.String(optionBeaconBase.Name)
+	// Initialize random seed
+	rand.Seed(time.Now().UnixNano())
+
+	// Load configuration
+
+	// Validate required configuration
+
+	log.Printf("[INIT] Starting blockchain indexer with StarRocks database")
```
aloknerurkar, line 143: In all other parts of the code we use slog. This outputs structured json logs which are easier to parse. It has option for text logging also and it can be configured. The other services also use this and we have some util pkg also. Better to start using this in the codebase.

All this manual `[COMPONENT]` can be added using log tags. This gives us the ability to filter using the tag in the logs instead of doing a text search.

tools/indexer/pkg/backfill/backfill.go
```diff
@@ -0,0 +1,250 @@
+// pkg/backfill/backfill.go
```
aloknerurkar, line 1: Any reason for this comment?

tools/indexer/cmd/indexer/main.go
```diff
@@ -0,0 +1,434 @@
...
+			log.Printf("[BACKFILL] Starting backfill operations...")
+			backfill.RunAll(ctx, db, httpc, createOptionsFromCLI(c), relays)
+
+		case <-mainTicker.C:
+			nextBN := lastBN + 1
+
+			// Fetch execution block data
+			ei, err := beacon.FetchCombinedBlockData(httpc, infuraRPC, beaconBase, nextBN)
+			if err != nil || ei == nil {
+				log.Printf("⏳ [BLOCK] Block %d not available yet: %v", nextBN, err)
```
aloknerurkar, line 236: Although I dont mind all these characters, not sure if anyone will be looking at this if things are running correctly. In which case you will only be looking at them when you have errors. Easier to parse the logs with lesser characters.

tools/indexer/cmd/indexer/main.go
```diff
@@ -0,0 +1,434 @@
...
+			log.Printf(" [SHUTDOWN] Graceful shutdown initiated: %v", ctx.Err())
+			if err := db.SaveLastBlockNumber(ctx, lastBN); err != nil {
+				log.Printf("[SHUTDOWN] Failed to save last block number: %v", err)
+			}
+			log.Printf("[SHUTDOWN] Indexer stopped at block %d", lastBN)
+			return nil
+
+		case <-backfillTicker.C:
+			log.Printf("[BACKFILL] Starting backfill operations...")
+			backfill.RunAll(ctx, db, httpc, createOptionsFromCLI(c), relays)
```
aloknerurkar, line 228: Why is ticker needed for backfill? This operation should only happen once. Also it can run in parallel with the main go routine so it should just be started before we enter this loop.

tools/indexer/pkg/ethereum/cleint.go
```diff
@@ -0,0 +1,191 @@
...
+	"io"
+	"net/http"
+	"strconv"
+	"strings"
+	"time"
+
+	httputil "github.com/primev/mev-commit/indexer/pkg/http"
+)
+
+const optInABIJSON = `[
```
aloknerurkar, line 20: I gave this comment in the last PR as well. We can use the generated client in `contracts-abi`. All it needs is ethereum client and it will handle all the data types.

tools/indexer/pkg/http/client.go
```diff
@@ -0,0 +1,82 @@
...
+				Timeout:   5 * time.Second,
+				KeepAlive: 30 * time.Second,
+			}).DialContext,
+			MaxIdleConns:        100,
+			IdleConnTimeout:     90 * time.Second,
+			TLSHandshakeTimeout: 5 * time.Second,
+		},
+	}
+}
+func FetchJSONWithRetry(ctx context.Context, httpc *http.Client, url string, out any, attempts int, baseDelay time.Duration) error {
```
aloknerurkar, line 28: If you want to provide these type of utility functions we cannot just assume this is doing the right thing. Unit tests should be added for this pkg. If not we can use some library that provides this functionality.

tools/indexer/pkg/database/starrock.go
```diff
@@ -0,0 +1,193 @@
...
+	if err := conn.PingContext(pingCtx); err != nil {
+		conn.Close()
+		return nil, fmt.Errorf("StarRocks ping failed: %v", err)
+	}
+
+	return &DB{Conn: conn}, nil
+
+}
+func (db *DB) Close() {
+	db.Conn.Close()
```
aloknerurkar, line 40: This doesnt return error?

Also if its the job of the DB to close this, we should not be exposing the conn directly outside.

tools/indexer/cmd/start.go
```diff
@@ -0,0 +1,284 @@
...
+								relayBids += len(batch)
+							}
+							cancel()
+							batch = batch[:0]
+						}
+					}
+				}
+
+				// final flush
+				if len(batch) > 0 {
```
aloknerurkar, line 208: If main context is cancelled, this insert will be called with cancelled context.

tools/indexer/cmd/start.go
```diff
@@ -0,0 +1,336 @@
...
+	)
+
+	if err := db.UpsertBlockFromExec(ctx, ei); err != nil {
+		logger.Error("[DB] failed to save block", "block", nextBN, "error", err)
+		return lastBN
+	}
+	logger.Info("[DB] block saved successfully", "block", nextBN)
+
+	processBidsForBlock(ctx, db, httpc, relays, ei, logger)
+	launchAsyncValidatorTasks(ctx, c, db, httpc, ei, beaconBase, logger)
```
aloknerurkar, line 144: We launch async tasks and save block progress? What if there was an error in the async task?

tools/indexer/cmd/start.go
```diff
@@ -0,0 +1,336 @@
...
+		logger.Error("[PROGRESS] failed to save block number", "block", blockNum, "error", err)
+	} else {
+		logger.Info("[PROGRESS] advanced to block", "block", blockNum)
+	}
+
+}
+
+func launchAsyncValidatorTasks(ctx context.Context, c *cli.Context, db *database.DB, httpc *retryablehttp.Client, ei *beacon.ExecInfo, beaconBase string, logger *slog.Logger) { // Async validator pubkey fetch
+	if ei.ProposerIdx != nil {
+		go func(slot int64, proposerIdx int64) {
```
aloknerurkar, line 236: So the first go routine is supposed to populate the Validator public key which the second routine is using? What is the point of doing it in separate go routines then? The second one will fail till first one completes. Not sure if I understand this logic.

tools/indexer/pkg/backfill/backfill.go
```diff
@@ -0,0 +1,171 @@
...
+			errCh <- err
+		}
+		close(slotChan)
+	}()
+
+	// Collect slots and validator data from channel
+	var slotsForBids []int64
+	var validatorsToCheck []SlotData
+
+	for data := range slotChan {
```
aloknerurkar, line 139: Better way is to have the slotChan be a channel of some type
```
type result struct {
  data SlotData
  err Error
}
```

This way you need only 1 for loop and you can check
```
if data.err != nil {
  // handle the error
}
```

tools/indexer/pkg/backfill/backfill.go
```diff
@@ -0,0 +1,171 @@
...
+		opted, err := ethereum.CallAreOptedInAtBlock(httpc.HTTPClient, cfg, v.BlockNumber, v.ValidatorPubkey)
+		if err != nil {
+			logger.Error("opt-in check failed", "slot", v.Slot, "error", err)
+			continue
+		}
+		if uerr := db.UpdateValidatorOptInStatus(ctx, v.Slot, opted); uerr != nil {
+			logger.Error("opt-in status update failed", "slot", v.Slot, "error", uerr)
+		}
+	}
+	if err := recentBids(ctx, db, httpc, relays, slotsForBids); err != nil {
```
aloknerurkar, line 163: So again, I am not sure you understood what I meant when I asked you to use channels. The idea was that we do it one by one, so get missing slot, then get the validator info and opted in status for that slot and then get the bids for that slot, then move onto the next slot.

You have just changed this validator check to be sequential but then the bids again use batches. This does not make any sense. Either you do one by one or do batch like you were doing before.

PR #803, opened 2025-09-24
feat: slash providers using slash amount    

shaspitz:
Updates provider registry to slash providers solely based on slash amount set in the commitment. Subsequent PR to mev-commit bidder node will set slash amount to bid amount as default, if slash amount is not specified.

### DIFF
--- a/.github/workflows/infrastructure.yml
+++ b/.github/workflows/infrastructure.yml
@@ -40,7 +40,7 @@ on:
           - chi1
           - mia2
           - mia3
-        default: 'lax'
+        default: 'lax1'
 
 permissions:
   contents: read


contracts/contracts/core/ProviderRegistry.sol
```diff
@@ -99,34 +99,26 @@ contract ProviderRegistry is
      * @dev Slash funds from the provider and send the slashed amount to the bidder.
      * @dev reenterancy not necessary but still putting here for precaution.
      * @dev Note we slash the funds taking into account the residual bid percent after decay.
-     * @param amt The amount of bid to slash from the provider's stake .
      * @param slashAmt The amount to slash from the provider's stake.
      * @param provider The address of the provider.
      * @param bidder The address to transfer the slashed funds to.
-     * @param residualBidPercentAfterDecay The residual bid percent after decay.
      */
     function slash(
```
shaspitz, line 106: Most important change in this PR is here

PR #820, opened 2025-10-03
fix: emit event for manually adding builder pubkey, introduce rm func

shaspitz:


contracts/contracts/core/ProviderRegistry.sol
```diff
@@ -219,6 +219,24 @@ contract ProviderRegistry is
...
+        bytes[] storage keys = eoaToBlsPubkeys[provider];
+        for (uint256 i = 0; i < keys.length; i++) {
+            if (keccak256(keys[i]) == keccak256(blsPublicKey)) {
+                keys[i] = keys[keys.length - 1];
+                keys.pop();
+                break;
+            }
+        }
+        delete blockBuilderBLSKeyToAddress[blsPublicKey];
+        emit BLSKeyRemoved(provider, blsPublicKey);
```
owen-eth, line 239: An event shouldn't be emitted if we don't find the key in the list, could require at the beginning that `blockBuilderBLSKeyToAddress[blsPublicKey]` is populated

owen-eth: Actually this could cause a different issue in the case of duplicates existing, since if we had this check then only one could be removed, and would fail on the second call. Maybe we want to consider removing all instances of the key?

contracts/contracts/core/ProviderRegistry.sol
```diff
@@ -219,6 +219,24 @@ contract ProviderRegistry is
...
+    function overrideRemoveBLSKey(
+        address provider,
+        bytes calldata blsPublicKey
+    ) external onlyOwner {
+        require(providerRegistered[provider], ProviderNotRegistered(provider));
+        bytes[] storage keys = eoaToBlsPubkeys[provider];
+        for (uint256 i = 0; i < keys.length; i++) {
+            if (keccak256(keys[i]) == keccak256(blsPublicKey)) {
+                keys[i] = keys[keys.length - 1];
+                keys.pop();
```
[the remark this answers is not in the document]
owen-eth: lgtm but maybe we want to remove all instances per comment below

contracts/contracts/core/ProviderRegistry.sol
shaspitz: I like the final idea here -> https://github.com/primev/mev-commit/commit/068285c3da778499aad55f03b9acbf86a34738fc

PR #829, opened 2025-10-13
fix: fast-track for RPC service

aloknerurkar:
## Describe your changes
Allow authenticated API to configure a provider set as a fast track group. If all the providers in the fast track group commit, we will preconfirm the transaction.

Fixes # (issue)

### DIFF
--- a/tools/preconf-rpc/service/service.go
+++ b/tools/preconf-rpc/service/service.go
@@ -38,6 +38,8 @@ import (
 	"google.golang.org/grpc/credentials"
 )
 
+const defaultMaxBodySize = 1 * 1024 * 1024 // 1 MB
+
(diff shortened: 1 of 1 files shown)


tools/preconf-rpc/main.go
```diff
@@ -203,6 +203,13 @@ var (
 		EnvVars: []string{"PRECONF_RPC_WEBHOOK_URLS"},
 	}
 
+	optionAuthToken = &cli.StringFlag{
```
shaspitz: Do we need a refresh interval of 1h for the auth token if it's only loaded on startup? Is the token intended to rotate every hour?

shaspitz (approved):

PR #831, opened 2025-10-15
perf: add cache for proxied read requests

aloknerurkar:
## Describe your changes
Adding cache for read methods not handled by the RPC. This is to reduce the load on the RPC service due to additional requests sent by wallet software.

Fixes # (issue)

### DIFF
--- a/tools/preconf-rpc/rpcserver/rpcserver.go
+++ b/tools/preconf-rpc/rpcserver/rpcserver.go
@@ -67,21 +67,18 @@ type jsonRPCError struct {
 }
 
 var cacheMethods = map[string]bool{
-	"eth_call":             true,
-	"eth_estimateGas":      true,
-	"eth_getCode":          true,
-	"eth_getStorageAt":     true,
-	"eth_feeHistory":       true,
-	"eth_gasPrice":         true,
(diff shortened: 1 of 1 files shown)


tools/preconf-rpc/rpcserver/rpcserver.go
```diff
@@ -187,44 +279,97 @@ func (s *JSONRPCServer) writeError(w http.ResponseWriter, id any, code int, mess
...
+		return 3 * time.Second
+	case "eth_call":
+		// if block tag provided and hex number → immutable
+		if strings.HasSuffix(string(params), "\"") { // cheap check
+			if strings.Contains(string(params), "\"0x") && !strings.Contains(string(params), "\"latest\"") {
+				return 24 * time.Hour
+			}
+		}
+		return 1 * time.Second
+	default:
```
shaspitz: `eth_getBlockByNumber` with 'latest' would hit this default case of 2sec. Could it be feasible to just not cache calls like this that are time sensitive? As-is we could delay a client's knowledge of the latest L1 block by up to 2 sec

shaspitz (approved):

